PR #30, opened 2015-12-30
Update header, change treatment of the downoad button to be a link in…

DanVanAtta:
…stead of a button. Change Documentation link copy to 'Help & Documentation'

[Review on Reviewable](https://reviewable.io/reviews/triplea-game/triplea-game.github.io/30)

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -7,15 +7,17 @@
 
         <li>
           <h1><a href="{{ "/" | prepend: site.baseurl }}">TripleA</h1></a></li>
+        
+        <li class="menu-items"><a href="http://github.com/triplea-game/triplea/releases/latest">Download</a></li>
+        
         <li class="menu-items"><a href="{{ "/contribute" | prepend: site.baseurl }}">Contribute</a></li>
 
         <li class="menu-items"><a href="{{ "/maps" | prepend: site.baseurl }}">Maps</a></li>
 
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Screenshot of hte updated header:
updated_header

williamdekryger:
Is there a general consensus on this chang?. I do like it, since I see the redundancy of two buttons, but the button implies that the file will begin downloading, and since that is what happens, this is not good for UI, as the link makes users think that they are getting a download page. How about we save this PR for when I find the time to work on the site a bit and write the download page, for which this will work nicely.

DanVanAtta:
I think if there is consensus between you and I, that is sufficient. Nobody is particularly watching the webpage, nor is anyone even visiting it. I think we should move relatively quickly with incremental benefits.

IT is a good point of a button vs a link. How about the download link sending people to the github releases page instead? Then when the download page is ready it'll be easy to change the URL. I would like to do some further work on the help and documentation page, thus I would like this change to be integrated sooner rather than later as well.

I'll assign to myself and make the link update. In the meantime, let me know if that is an agreeable way forward, updating the link to github releases page and then merging.

DanVanAtta:
getting rid of the extra red box is overall better for the UX, even if it is not perfect. Merging.

DanVanAtta:
@williamdekryger , actually, the header link opens up the releases page, it makes sense. The remaining download button likely should be updated, but that wasn't the subject of this PR.

PR #35, opened 2016-02-25
Added CNAME for new domain

williamdekryger:
Merge this when the DNS records start working.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+triplea-game.org


DanVanAtta:
let's give the DNS a try, it's been 2 days

DanVanAtta:
@williamdekryger , not working immediately. Perhaps something is still wrong. We could give it some time in case something needs to update somewhere..

PR #39, opened 2016-03-16
Added Download Page

williamdekryger:
I have completed the download page in an early simple form. Here is a screenshot:

screen shot 2016-03-16 at 8 06 07 am

Clicking one of the buttons begins the download in the background while displaying installation information below.


gaborbernat:
it has merge conflicts, can you look on those? :)

williamdekryger:
Yeah. I'll deal with those. BTW, there is a bit more work I would like to do before merging, but I wanted to show off some of the updates I had been working on.

williamdekryger:
@DanVanAtta To your earlier (a few months ago) point, `_config.yml` will need to be updated every new version with new links and a new version number. This is detailed at the bottom of the README.

DanVanAtta:
@williamdekryger  I would  drop the "please select your operating.." text completely. It should be obvious that a person should click on their OS. From there, the steps should also be obvious, so no need to detail the full deployment process up front to the user, they should be able to follow the flow with relatively minimal prompts / instructions.

The install powered by I think would be better a bit less prominent, below the next line and perhaps right justified. This last point I don't feel super strongly about, so more advistory.

PR #61, opened 2016-07-10
Revert "Revert "Added Links | Fixed Markdown""

RoiEXLab:
This Reverts the revert of my PR and fixes the error... 2 simple quotes were missing...


RoiEXLab:
@DanVanAtta the reason for the config change, was due to the FB thingy. Facebook displayed &colon; instead of `:` If it's now displaying quotes, then this won't really be noticed

PR #64, opened 2016-07-11
Added slideshow

RoiEXLab:
I added a slideshow instead of the "windows XP image collection".
Works, but is not finished, since the images provided on the server aro too low res...
I will add images with higher resolution tomorrow... (Or @DanVanAtta merges it and somebody adds the images later on)
Also since the images have different sizes, we have to adjust the vertical reserved image size.
screenshot 1
screenshot 2
screenshot 3


DanVanAtta:
What do you think of adding the original homepage image as the first slide?

RoiEXLab:
@DanVanAtta updated...
Images are now taken in high resolutions...
We should decrease the Image size on some images with huge file sizes (e.g. 10MB) and crop them to a better fitting format...

I added a Image similar to the previous homepage image as first image... tell me what you think about it...

index.html
```diff
@@ -22,17 +22,37 @@
...
           <div class="text">Big World</div>
         </div>
         <div class="mySlides fade">
-          <img src="./images/maps/270BC.png">
+          <img src="./images/maps/high-res/270BC.png">
           <div class="text">270BC</div>
         </div>
         <div class="mySlides fade">
-          <img src="./images/maps/GreatWar.png">
+          <img src="./images/maps/high-res/ClassicWW2-3rd.png">
```
DanVanAtta: Avoid committing binary files to github. Use the github file hosting service to do this instead. Open a new issue, upload an image, and use the link it creates. You do not even need to submit the image.

DanVanAtta:
I'm concerned about page load times with all the images. Can you speak to any impacts the images will have? Have you run any tools like 'ySlow' to make sure that the page rendering does not block while downloading images?

Ultimately, I think 10MB for an image on a website is a deal breaker, it is just too large to have. This website will barely load if someone is working over a slow connection (which is still going to be very common in many parts of the world, including large sections of the USA which still has a lot of people on 56k connections even)

DanVanAtta:
With the original image back in the slide show, functionally I do not have objections, this seems like a good improvement. I think we may need to deal with the image size first though before this can be merged, please see my comment above.

RoiEXLab:
@DanVanAtta I decreased the Image size to a maximum of 5 MB each...
Resolution is mostly to around 2K - FullHD.
I'm probably going to replace the current image with a "newer one" - which looks exactly the same, but newer...
Also I added the requested image back in if you didn't tell from the context

DanVanAtta:
@RoiEXLab , thanks for your work on this. Please remove the images from Git, ie: upload them to github via the issue image upload, and then use the URL given to you by github. Binary files should not be checked in to version control. Just that last thing and this otherwise looks good.

RoiEXLab:
greatwar
lotrmiddleearth
napoleanicempires
waw
270bc
bigworld
classicww2-3rd
triplea


@DanVanAtta updated!

DanVanAtta:
@RoiEXLab  during the transition the image fades to white. It makes it hard to focus on any other part of the page. Can this flash be reduced? IE: at some point a person should be able to ignore the slide show when they wan to,

RoiEXLab:
@DanVanAtta sure - I forgot to remove this... 
I'm also going to decrease the placeholder size for mobile devices

PR #83, opened 2016-08-29
Update the Download Page for 1.9

DanVanAtta:
- Split windows download button between 32 bit and 64 bit
- Clean up download text in the download button
- Remove the prerelease links- Clean up the installation instructions, update the linux instructions, add help link to mac, prune down the windows section
- Clean up and update the installation instructions

Additional Changes:
- Remove direct download direct, if the download buttons do not work, the releases page will not be of any further help.
- Move the install4j credits to the top right and make the text smaller
- Move the rulebook and release notes to the OS specific instruction sections, the goal here being to unclutter the initial view and not have any links below the download buttons.

Before:
before

After:
after

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -7,6 +7,7 @@ markdown: kramdown
 
 # TripleA version information. Update each time a new version is released.
 triplea-version: 1.8.0.9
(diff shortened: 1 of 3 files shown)


DanVanAtta:
PRs to improve the download page further are very welcome!!

Issue #104, opened 2016-09-06
Download links point to 3089 instead of 3096


panther2:
#103 edited 3089 in, instead of 3096

panther2:
Fixed and closed.

PR #105, opened 2016-09-06
Correct links to version 3096

panther2:
This solves issue #104


DanVanAtta:
Thanks for the patch. It does seem in the future we should move the version to a variable, instead of having the full URL be a variable. Then we would be able to change a single value instead of 9.

panther2:
My pleasure. Yes, that sounds good.

Issue #97, opened 2016-09-03
CSS default to show bullets


DanVanAtta:
I was looking for where we turn off bullets on the list items by default. Where is that config? 

Second, I've frequently have wanted bullets, thoughts on flipping it so bullets are on by default for `<li>` and `<ol>`  items?

For example, this markdown page: http://www.triplea-game.org/bug_report/, does not look pretty, does not seem ideal that we have to special case it to make it look like a list. This then lends into that we use `<p>` tags too much, around every sentence, having lists be displayed by default would cut down on that.

### Example

Code, looks like a list:
![code](https://cloud.githubusercontent.com/assets/12397753/18227324/be387a00-71d5-11e6-9501-ec6c369804c7.png)

Default output, formatting is not clean:
![view](https://cloud.githubusercontent.com/assets/12397753/18227325/c1d73a34-71d5-11e6-83c8-f23e793f1f31.png)

DanVanAtta:
Now done, thank you @williamdekryger

PR #111, opened 2016-09-07
Added permalink to Upgrade Notes and Fix Bullet Points

williamdekryger:
All pages on the site should have a permalink in the YAML front matter. This, according to Jekyll URL logic should be throwing a 404, and does on my local copy, though weirdly enough, it doesn't on the live site. Anyways, let's add this to be sure it doesn't give us problems in the future.

This pull also addresses #97, changing all `<ol>` and `<ul>` elements to `display: list-item`, restoring bullet point display.


RoiEXLab:
No objections on merging...

PR #114, opened 2016-09-08
Basic Maps Page

williamdekryger:
An update to the maps page to show all maps, pulled directly from `triplea_maps.yaml`.

Adds five pages to the site: a "featured" maps page, which for now, shows the "best" category of maps, a page with all maps sorted by category, a mods and skins page, a map tools page, and a page with all the maps in the order they appear.

Here is a screenshot of how it looks currently:

Maps Page Screenshot

### Roadmap

* [ ] Work bugs out of the current five pages
* [ ] Finalize new requirements for map-makers
* [ ] Add eras as a method for sorting maps
* [ ] Update current maps with new requirements in `triplea_maps.yaml`
* [ ] Add new information to the maps page
* [ ] Merge

Work will continue to take place in the `maps_page` branch.


RoiEXLab:
I think closing this PR and opening another one should get rid of the large diff... If your fork is properly rebased of course

williamdekryger:
My fork is properly rebased. It is the `maps_page` branch that is not rebased. Someone with repo rights could do that, and I'd resend the PR.

RoiEXLab:
Ah I see... Sorry mobile view doesn't shows that...

DanVanAtta:
yes, the diff is a bit difficult. It appears first the maps-repo branch needs to be rebased or merged to latest master. AFter that, this diff should not be so large.

DanVanAtta:
Question, why do we have a download button? 
- we have an in-game mechanism to download, should we not be encouraging players to use that? IMO the biggest benefit of the maps page is to showcase the maps, not necessarily to act as a portal for downloading them. If we allow download, we'll need instructions for players on how to install. I don't think I want people to be 'trained' to download maps from here. IMO replacing the download to a link to the github repo page woudl be a good move. It does seem we plan to use the README of those repo's, in which case there will be further information available. Also the repo's have a download button as well.

DanVanAtta:
Second, IMO remove the map tools. I *really* want to kill that category. It's cludgy/hacky to bootstrap the map tools downloads with maps. That was something clearly done just because it was convenient, and cheaper than doing it in a more robust manner. The map tools I expect to disappear from the DL list. They are different enough they'll want their own documentation+download page.

DanVanAtta:
Last point, I think it might be most effective to focus first on getting a basic page out:
- map title that links to the repo
- map thumbnail
- description

From there I think we should then ask questions of what information do we *really* want to see on this page. IT should then be pretty easy to shim additional data points in. Any such data point we need to consider really carefully. Looks like we have about 5 new ones being proposed, with about 100 maps, that is 500 data points for us to hand manage! WE had better be sure we are ready to take on that overhead, and really benefit from it.

williamdekryger:
@DanVanAtta I'm fine with getting rid of the download buttons for now. I'll have another suggestion to make on that later 😉 

I agree with removing the map tools. Let me know when you want me to give them their own page. They'll stay with the maps until we find another solution.

I also like trying to get out a basic maps page soon. For now, I think we'll end up linking the download buttons to a page explaining how to download maps in-game.

DanVanAtta:
@williamdekryger , what do you think of moving these updates to a new file? That way only if you know the URL can you see the page, that would help us preview it and allow for you to iterate on the code more.

williamdekryger:
Alright, I've added a few changes, and I believe that this is not only ready for merge into maps_page, but maybe even into master, so we can begin testing. I've put the maps page at the url http://triplea-game.org/maps-list/ if it were live. That way the regular page can stay up until we finalize the new one.

I've also copied the roadmap into the corresponding issue, so we don't have to refer to this PR anymore.

williamdekryger:
@DanVanAtta Any objections to merging this? It would be completely experimental, and nothing would link to it. The URL of the page (when it does get merged) is above. If there are no objections, do you mind merging?

Just as a note, I changed the base of the PR to master, rather than maps_page, so we can get the page up right away. Once we like our progress, it's fairly simple to delete the old maps page and change the URL to "maps."

DanVanAtta:
Ah, I thought this was something slightly different, yeah, no objections.

PR #120, opened 2016-09-25
'Old downloads' page

DanVanAtta:
- moved around links on download page to bottom right and added a link for the 'old downloads' page
- added an 'old downloads' page
- simplified the upgrade notes

Download page before:
before

Download page after:
after

New 'old downloads' page:
old_downloads


DanVanAtta:
Feedback is welcome!! Please post any comments and/or submit PRs .

PR #156, opened 2017-01-05
Added dynamic Download links

RoiEXLab:
Also the filesize adjusts when hovering over the installer
Links inside the a tags are kept for clients with non working or disabled javascript.
In theory we no longer need to edit the download page on every new triplea release


RoiEXLab:
If you didn't guess already this uses the github api, the script gets the first non-prerelease release out of every triplea-release ever and sets the links accordingly

RoiEXLab:
@DanVanAtta Your thoughts on this change?

DanVanAtta:
This is very interesting. I think we might even be able to do the latest prerelease version, then cut out the manual step of updating the release type. 

It would be great if we could remove the need to update the manual fallback. It would also be nice to remove the GET request from page load as well. Static HTML content is very fast HTML content. 

This additional functionality I think can potentially be a good iteration. IMO longer term we may be able to remove all of the manual update by baking this update into the travis build.

RoiEXLab:
I mean in theory every modern browser supports this kind of javascript, so we could easily set this as a requirement... 
The bigger Problem I saw is the potential effect of plugins like noscript etc.
Haven't tested those with my code...


I think the idea of adding a script to the travis build script is not bad, but the question I can't answer myself is how we are going to edit the config.yml via buildscript since we don't want to update the buildscript every time we want to update this file.
Since Linux provides us tools like curl or similar and I believe we can use python or similar languages using travis the other part shouldn't be a problem...

RoiEXLab:
@DanVanAtta I tried to find a good method to achieve the expected result...
All of my attempts require a local clone of the gh pages Repository followed by a push...
The First option would be to edit the config.yml file using python or ruby and then re-set the version value, adding 4 filesize values and adding those values in the html file to being processed with javascript.
Downside of this would be, that the comments inside the config file would be removed due to the yml apis.
The other attempt would be the same, but instead we would replace the values using regex which would keep the comments.
We could also add the direct links to the config file to have them work even when we changed the file naming.

The only problem I see, is that we can't tell if we are talking about a prerelease or a full release. 
If we want to release something we let travis run and deploy as usual, and then unset the prerelease checkbox, but thats AFTER deployment, the buildscript will still treat this as a prerelease. The only way to get that working is to repush the repository after unchecking the checkbox or to edit the travis file before a release.
For now I'd like you to merge this PR until we find a good solution...

RoiEXLab:
@DanVanAtta Should we merge this as non-required additional solution?
Or should I close this until we come up with a better solution?

DanVanAtta:
Perhaps close. I like the direction of the update, but since we still will be updating the fallback value, there is no practical benefit just yet.

DanVanAtta:
@RoiEXLab - just realized - the dynamic game size I think we could use right away. That would avoid having to update that value by hand. If the current game version does not match the latest pre-release, that is a mistake already on our part, but also a very minor one since the download size is a ballpark, and it being off by a little bit is okay.

RoiEXLab:
Reopening then... I'll remove the "unused" part on Tuesday

RoiEXLab:
@DanVanAtta I think I found a solution for our problem...
Since Github provides webhooks in order to call a server side script, we could setup a script like [this one](https://gist.github.com/RoiEXLab/92ba917e658666a26bc41f6607ccc825) on a PHP server like warclub, which updates the website every time the webhook is being fired.
Unfortunately there is no such event, which fires whenever the prerelease status of a release is changed.
I asked the Github support:
> That's correct -- there's no event for this currently. I'll pass your feedback to the team to consider, but I can't say when such an event might be added.

That would of course be the best option, but there's an alternative...
If @DanVanAtta or @ron-murhammer would create a webhook, listening to pushes, the script I created checks if this push contains no commit (hopefully - haven't tested it yet), and then checks if triplea has released a new full release.
If so, the config file of this repository is downloaded and changed and uploaded again with updated version and size values. This removes all comments in this file...
Downsides:

- We need a PHP server
- We need to install the PHP cURL extension
- We need to install the PHP yaml symfony component extension
- You need to push to the triplea repository every time a new major version is released in order to update the website (won't be the case any longer, once the new feature has been added)
Opinions?
Of course we would need to update this website a single time once this feature was added.

DanVanAtta:
I do wonder if this would be a simpler approach:
- in the travis build file, as a post build step
- clone the website repository
- update the version number
- commit and push it

Thoughts @RoiEXLab ? I think the only real problem above is that sometimes the push to github fails (we're pushing about 2 or 3 gigs per release, so not terribly surprising). We might be able to add some error checkign in to make sure we only update the version if the release was successful.

RoiEXLab:
@DanVanAtta correct me if I'm wrong, but I believe the current way to release a stable version of triplea is to change the status from prerelease to release...
If this change is done afterwards, the buildscript has no way to get this information which will be done in the future... 
Thats why the buildscript attempt won't work

DanVanAtta:
I don't think anything depends on the prerelease flag, updating it is a bit more of a formality/convenience so that the /latest URL works.

We can update it so each travis build is a release as well:
https://github.com/triplea-game/triplea/blob/master/.travis.yml#L60

RoiEXLab:
@DanVanAtta The point is, that I believe it's not the best idea o update the website every time triplea is built on master...

RoiEXLab:
> I don't think anything depends on the prerelease flag, updating it is a bit more of a formality/convenience so that the /latest URL works.

Should have noticed that earlier... This makes getting the latest release much easier ^^
Updated my PHP script to match that...

@DanVanAtta simple question: When is the website supposed to be updated?

RoiEXLab:
@DanVanAtta another Idea... We could require javascript, and have the [latest link](https://github.com/triplea-game/triplea/releases/latest) as fallback if js is disabled or not available... This would get rid of the GET request when the site is loading, but since we're only downloading a little bit of json taht probably won't even matter...

RoiEXLab:
Ok implemented the functionality this way...
@DanVanAtta thoughts? There's no longer a need for updating the config.yml file...

RoiEXLab:
Sorry @DanVanAtta I couldn help myself, I needed to fix some javascript on this download site

js/scripts.js
```diff
@@ -1,11 +1 @@
...
-    return;
-  }
-  if (which.style.display=="block") {
-    which.style.display="none";
-  }
-  else {
-    which.style.display="block";
-  }
-}
+//Nothing here
```
RoiEXLab: I will remove this file, once my other penging PR is merged.
Otherwhise we would have to handle merging conflicts

RoiEXLab:
@DanVanAtta if you aren't convinced, I tested the latency on my local branch... It took 300ms to perform the GET request downloading about 13KB, considering my Internet connection is pretty bad (16K minimum ping is 31ms) that's a result that shouldn't really have a huge impact on the loading experience since its performed in the background

RoiEXLab:
@DanVanAtta I removed the unused script file and resolved all merge conflicts...
Make sure to uncheck the prerelease status of the [current release](https://github.com/triplea-game/triplea/releases/tag/1.9.0.0.3627)

RoiEXLab:
@DanVanAtta thoughts?

RoiEXLab:
@DanVanAtta bump

RoiEXLab:
@DanVanAtta please merge this PR...

DanVanAtta:
Sorry @RoiEXLab  this slipped through the cracks a bit.

Question: if we do have a problem with the latest build, will we be able to pin the release version to a previous one while we fix the latest build?

RoiEXLab:
This PR always picks the last triplea version marked as "full-release" (not pre-release).
So yeah. That can be changed in gh

DanVanAtta:
@RoiEXLab  - idea for this that struck me. If we introduce a bug we will want to hold back the release version on the website, but otherwise we want it to keep marching along (or until we change compat level, but ideally that is done very explicitly and in a coorddinated way).

So, I suggest allowing for a manual override we can set when we want to pin the version. Otherwise the website goes and uses the very latest version. Thoughts?

RoiEXLab:
This is the point.
It chooses the last build artefact which is marked as full release. (The one found using https://github.com/triplea-game/triplea/releases/latest)
We can change that using github, but we no longer need to change the website's code

DanVanAtta:
Ah, I see @RoiEXLab 
That sounds reasonable enough.
Would you mind fixing the rebase conflicts?

RoiEXLab:
@DanVanAtta There are none?

RoiEXLab:
@DanVanAtta @ron-murhammer Would you mind merging this change?

RoiEXLab:
@ron-murhammer 
@DanVanAtta already approved this change...
This PR was opened in January! I believe it's time to finally merge it!

ron-murhammer:
@RoiEXLab Can you summarize this PR? It looks like you and Dan had a fairly lengthy review which didn't end up with it being merged.

RoiEXLab:
@ron-murhammer This PR has gotten a bit messy since it was rebased once.
Basically what it does is that instead of "generating" download based on the version number set inside the config, it uses a fallback link which is overriden once the javascript snippet figures out the latest stable release of triplea using the github API.
@DanVanAtta Already approved this I think, but didn't merge because of some non-existent Conflicts I guess?

DanVanAtta:
conflicts


 : )

This is strange that you're not seeing conflicts. FWIW git hub has these command line suggestions:
```
Step 1: From your project repository, check out a new branch and test the changes.

git checkout -b RoiEXLab-master master
git pull https://github.com/RoiEXLab/triplea-game.github.io.git master
Step 2: Merge the changes and update on GitHub.

git checkout master
git merge --no-ff RoiEXLab-master
git push origin master
```

DanVanAtta:
@RoiEXLab  I think we need to be sure to clarify requirements and needs. I apologize for the churn so far, coming to a good and needed solution is valuable. I actually suspect we may also want to enhance this is a bit.

Based on your last explanation, I also think my understanding of where this is, was incorrect. To clarify, the two behaviors I think we could really use:

1. Check dynamically for the latest version, get that URL and supply that as the download URL (with the hardcoded default just in case perhaps.). Thinking about this one a bit more though, linking to the download page is probably better compared to giving a non-latest version. It is debatable though, and I think depends on how out of date the hardcoded version woudl get.

2. In case our latest version is bad, for whatever reason, we need a manual override to 'pin' the latest version to a specific version. Ideally we never use this, but in case we find in testing that the latest is so not backwards compatible, we would want to turn this on until things are fixed.

 @RoiEXLab  I do not believe we have this second item coded up, just the first? I'm happy to clarify/discuss further. What are your thoughts on the above otherwise @RoiEXLab , reasonable?

ssoloff:
@DanVanAtta Weird... as @RoiEXLab observed, I'm not seeing any conflicts either via the GitHub UI:

issue-156-conflicts

RoiEXLab:
I made a merging mistake...
Don't merge anything until I force reset this mess

RoiEXLab:
Replaced by #180

PR #160, opened 2017-01-15
Add git setup and tips faq section with some notes on rebasing and me…

DanVanAtta:
…rging

- TODO: git setup tips can be consolidated, likely some redundant info is being added.


RoiEXLab:
@DanVanAtta Looks good, can't confirm if those commands are correct, since I'm not using git via command line ^^

Issue #33, opened 2016-02-23
Maps Page


theredbaronaaa:
I am creating this thread as a place for discussion on the map page that I am currently working on. The project will potentially be the only dynamic content on the site, with the JSON features of Javascript being used to edit the DOM upon page load. A few things have already been discussed in #11.
### Current Features being worked on
- Uses the Github API to access map info dynamically from [here](https://api.github.com/orgs/triplea-maps/repos)
- Because of the above, we can display how many Github stars each map has and provide a direct download link.
### Problems/Questions
- The Github API can only be called once a minute, so we need to potentially script something to copy it to another location every few hours.
- Thumbnails, author information, and other things need to be standardized so that I can call them based on the urls I have from the API. Perhaps, they could all be placed in the map folder as readme.txt and thumbnail.png. Ideas for standards for those?

I will soon be pushing to a maps_page branch. I welcome and any ideas or comments you leave below.

Here is a screenshot of how it looks currently:

![Maps Page Screenshot](https://cloud.githubusercontent.com/assets/5250014/18347839/697467e6-75b7-11e6-9073-6e074a373b84.png)
### Roadmap
- [x] Get basic maps page working
- [ ] Work bugs out of the current five pages
- [ ] Finalize new requirements for map-makers
- [ ] Add eras as a method for sorting maps
- [ ] Update current maps with new requirements in `triplea_maps.yaml`
- [ ] Add new information to the maps page
- [ ] Merge

Work will continue to take place in the `maps_page` branch.

DanVanAtta:
One note, the github API automatically pages the response, need to add some args to see the full list:
`https://api.github.com/orgs/triplea-maps/repos?page=1&per_page=10000`

The API rate limit is an interesting problem. Another solution would be to regenerate the static content and republish every now and then. Thinking along those lines, we could add a deploy hook to each map to invoke a script that will regenerate the site content.  Then each time the maps are updated, so will the site.

williamdekryger:
Interestingly, I just found out that Jekyll supports "[Data Files](http://jekyllrb.com/docs/datafiles/)," which basically would allow us to put the JSON file in the `_data` folder and call it into a page via the `site.data` variable. This seems way easier than coding a DOM script.

williamdekryger:
[...]

PR #174, opened 2017-04-02
Update maps.md

panther2:
... to remove as well outdated Sourceforge-links that lead to maps not compatible with version 1.9 as "comes with TripleA by default"-hints.

I am aware of #33 - but these corrections seem to be necessary in the meantime due to compatibility reasons. Also maps don't come "by default" any longer.


panther2:
@DanVanAtta 
@ron-murhammer

ron-murhammer:
@panther2 Looks good. Thanks for the updates. Merging now.

PR #178, opened 2017-04-10
Correct Dev. link to new forum.

prastle:
Removed the link to the old Development Forum /Source Forge & replaced with the new forum link.


prastle:
good point

PR #184, opened 2017-05-21
Html cleanup

RoiEXLab:
This fixes a lot of invalid html


RoiEXLab:
A lot of this PR was done using regex based find-and-replacing.
I noticed, that the big "documentation pages" are partly way outdated and partly broken (in terms of missing html)
I added some code highlighting using highlight.js and removed those "Go back to Top" href links and replaced them with a js based button for all pages which looks a lot nicer.
Other than that there are no functional changes. Just general refactoring and clean up.

PR #186, opened 2017-05-22
Collection change

RoiEXLab:
This removes the triplea_maps.yaml file and replaces it with a collection of documents.
This has the great benefit that we are now able to "auto-generate" a page for each map.
Every map is accessible at `triplea-game.org/map/:slug` Where slug is a trimmed, lowercase version of the map name where spaces are replaced by hyphens.
The files are autogenerated by a python script which needs to be installed on the triplea repo RIGHT AFTER THIS PR IS MERGED!
https://github.com/triplea-game/triplea/pull/1740


theredbaronaaa:
I briefly skimmed these changes but I must say that this is a very nice improvement. I'll have to pull it down and play with it in a bit 😉 

I'm going to hit pause on my updates to the map page while you get this new system working, so that I don't get everything set up before it gets changed. Also, you might consider changing the permalinks from "maps-page" to "maps" so that it gets linked to from the menu. It will be a bit rough for a little while, but not nearly as bad as what we currently have.

Also, the HTML that the maps use is really ugly. Can the engine not read `<ul>` or `<p>` instead of these hackish `<br>` tags? I could volunteer to reformat them, but I need to know that the engine could read them in that state.

RoiEXLab:
I agree, the html should be better.
The engine should understand HTML 4.01 but not HTML 5.
If triplea uses another framework one day we'll be able to use html5.

RoiEXLab:
After this PR gets merged I'll leave it up to you to design the map page if you want... 
I just realised that the current template is not perfect. I didn't realise there were also map skins in the yaml file so I didn't add a special case. Now they look like normal maps without a category but whatever ^^
Which permalink do you mean?

theredbaronaaa:
@RoiEXLab I'm talking about the difference between the [old maps page](http://www.triplea-game.org/maps/) and the [new maps page](http://www.triplea-game.org/maps-list/). Right now, the latter is not linked by any page and simply exists for testing purposes, thus the permalink outside of /maps, which will be the permalink once the page is finalized.

Therefore, I was thinking it might be a good idea to get rid of [maps.md](https://github.com/triplea-game/triplea-game.github.io/blob/master/maps.md) and push the listing into production. Then you would have to do a quick find and replace for "maps-list" and make it "maps," and then you would have the new page linked from the menu at the current location of the [old page](http://www.triplea-game.org/maps/).

On second thought, maybe I can make this change once this PR is merged, since then I can also patch up _layouts/map.html. Your thoughts on this?

RoiEXLab:
Ah ok...
Yes, I believe this would be better.
Once the maps page looks good, we can replace the current one. I'll leave this up to you then.

RoiEXLab:
I wrote a post about this in the forum: https://forums.triplea-game.org/topic/117
With some luck some map makers update their description on their own!

RoiEXLab:
I updated the layout page to be ready to be used in production.
@theredbaronaaa 
I'm not planning on editing it any further until this PR gets merged, so if you want to improve its design feel free to play around with this branch if you like \:-)

ron-murhammer:
@RoiEXLab @theredbaronaaa FYI, swing displays html kind of funny sometimes. So if you start redoing the html, you'll want to still test how it appears in TripleA map selection as well.

RoiEXLab:
@ron-murhammer I know, I hope everything works when just using HTML 4.01 without any _new_ HTML5 elements.
That's why I suggested markdown. It generates just p ol il li and a tags most of the time!
If we are going to move to another graphic framework like JavaFX one day this shouldn't be a problem any longer.

_layouts/map.html
```diff
@@ -0,0 +1,13 @@
+---
+layout: default
+---
+<h2>{{ page.mapName }} v{{ page.version }}</h2>
+{% if page.mapCategory %}<h4>Category: {{ page.mapCategory }}</h4>{% endif %}
+{% if page.mapType == "MAP_SKIN" %}<h4>Type: Map Skin</h4>{% endif %}
+<img src="{{ page.img }}">
```
DanVanAtta: I wonder if we should also have `page.img` inside an if block. 

Check markup counts in the maps yaml, we do have some data quality issues:
```
$ egrep -o "^  [a-z].*:" triplea_maps.yaml | sed 's/:.*//' | sort | uniq -c
    106   description
     93   img
     90   mapCategory
      1   mapName
     11   mapType
    105   url
    110   version
```
For some maps we will not have image it looks like. I think the options are to either always enforce an image, or put in tolerance for not having it. In the case of not having it, excluding altogether or using a default "no image" image would be some options.

ssoloff: I guess the schema I wrote for `triplea_maps.yaml` allows duplicate properties. 😞  I will investigate improving that...

DanVanAtta: More accurate grepping, 106 total entries:
```
$ grep -c "^-" triplea_maps.yaml 
106

$ sed 's/^-/ /' triplea_maps.yaml | egrep -o "^  [a-z]*:" | sort | uniq -c
    106   description:
     93   img:
    106   url:
    110   version:
```

RoiEXLab:
Rebased against current master.
@DanVanAtta 
> Smaller notes - Do you have strong attachments to hyphens instead of underscores? Using underscores could simplify, we could construct the download URL directly, and would avoid needing dlurl.

Unfortunately this wouldn't work because not all repositories use underscores as their repo names.
Example: https://github.com/triplea-maps/1914-cow-empires
Also I like decoupling names from actual URLs.
The files are generated automatically, so having an extra param has no downside IMO.
> Does the automation create new files if they do not exist yet? That would be a shortcoming when adding new maps. (side note, NBD: Might make sense to have the build automation create the split map files for us instead of including them here)

Of course. When writing to files in python they need to exist, or we'll get an exception thrown or you need to explicitly create the files [which I did](https://github.com/RoiEXLab/triplea/blob/ffd4db0a4de635069eefb9e9dbcf5cd96cbcc492/.travis/yaml_splitter#L12), because I don't have the time to create 100 of files manually \:P

RoiEXLab:
Because I'm a terrible designer I used a modified version of frostions map image we use in the forums in order to indicate "missing images".
We want every map to have one, but we shouldn't require it for now...
How it currently looks:
Screenshot

RoiEXLab:
@DanVanAtta @ron-murhammer Merge Conflicts resolved

DanVanAtta:
@RoiEXLab one more set of conflicts if you would please.

RoiEXLab:
@DanVanAtta 
Screenshot
Github is weird -.-
What conflicts are you seeing?

ssoloff:
@DanVanAtta @RoiEXLab This is the second time this has happened (I'm also seeing the same message @RoiEXLab is seeing).  I have a theory, though...

I believe @DanVanAtta is trying to rebase this branch onto `master` instead of merging it.  It looks like @RoiEXLab has already merged changes from `master` onto this branch instead of rebasing (e.g. 41a099c).  That's possibly preventing the rebase from succeeding.

I've found the following workflow to work as expected when I want to resolve conflicts in a branch without precluding rebasing back to `master`.  Assume the feature branch name is `my-feature` and it has gone out-of-sync with `master`, then (using Git 2.9.4):

1. `$ git checkout master`
1. `$ git fetch --all`
1. `$ git pull --rebase upstream master` # should fast-forward latest changes from upstream
1. `$ git push` # updates remote `master` on `origin`
1. `$ git checkout my-feature`
1. `$ git rebase -p master` # may have to fix conflicts during the rebase; -p ensures merge commits on `master` are preserved on `my-feature`
1. `$ git push --force` # updates remote `my-feature` on `origin`

where the `upstream` repo is the official TripleA repo.

DanVanAtta:
Thanks for jumping in.  
can_be_changed
^ that drop down can actually be changed. Changing the merge strategy does change the conflicts..

```
Rebase and merge -> conflicts
Squash and merge -> no conflicts
merge commit -> no conflicts
```
For force pushes, `git push --force `  it can be good to avoid it when a review of a branch has started. The history overwrite can cause commentary to get orphaned, and depending on circumstances some re-review is needed. Not to get caught up in the weeds, I think we're okay -  squash and merge is a good option.

RoiEXLab:
Thanks @ssoloff , this seemed to work...
The last command was actually different... merging detached the HEAD and I needed to type `git push origin HEAD:my-feature --force` instead...
@DanVanAtta current state?

RoiEXLab:
@ssoloff Nevermind... Something went wrong leaving my local repo in a state with detached head...
I usually don't use git by command line ^^
But I got it working again without having to reclone everything...
Should now work as expected @DanVanAtta

RoiEXLab:
Just for reference:
Aparrently this issue also occurs when merging via the github online interface...

ssoloff:
@RoiEXLab Sorry, I should have elaborated on the rebase step...

If there are no conflicts during the rebase, you are good to go.

If there are conflicts, you will be in a detached HEAD state (as you observed) until you conclude the rebase.  After resolving each conflicted commit, you need to continue using `git rebase --continue`.  You will ultimately rebase the final commit and your HEAD will be re-attached.  When you run `git status` during a rebase, you should see some helpful information at the top instructing you how to proceed or  possibly abort the rebase.

RoiEXLab:
@DanVanAtta After fiddling around with git everything should work fine now...
Rebasing locally works without any merge conflicts

DanVanAtta:
>> Does the automation create new files if they do not exist yet? 

I should rephrase perhaps. We expect the yaml splitter to create files automatically, and to publish new files. In theory it can create all the auto-generated files committed here from a clean slate. I'm thinking we would want to do that. It tests out the automation very nicely.

> Also I like decoupling names from actual URLs.

We're on the same page here. But, the problem is we have 3 things that all equivalent (1 thing), and instead we need to represent 2 things. To explain, we have `mapName` which is used as the map title and is also used as an input to a set of normalization functions that are used to find map folders on disk. Those map folders are written according to the download URL. The download url has a github slug repo bake into it, and while everything is in the same org, the url and slug are equivalent. The only freedom is that casing and white space are normalized away in the `mapName`, otherwise it has to match the slug name.

Instead I think we need:
- `title`
- `repo`

Maps hosted outside of triplea-maps I think can be handled by a new feature to download and install maps from user defined locations. Hence repo would always refer to the triplea-maps org. Title is what is shown when viewing the download. The title from the XMLs are shown the rest of the time.

TBH - there is quite a bit there to bite off. Updating triplea_maps.yaml to simplify it is not easy since it is a live configuration and has to be backwards compatible. I suggest for now we keep triplea_maps.yaml stable, but keep in mind that we will have new fields coming along at some point.

DanVanAtta:
@RoiEXLab summarizing some options for next steps:
- I think it's worthwhile to remove the auto-generated files, let the build automation create them. There is a bit of test value there, as well as building our confidence in the build automation
- consider collapsing redundant variables together (slug + url). The above has some future context. If you're happy with this iteration, I do request renaming `dlurl` to `downloadUrl` : )

_layouts/map.html
ssoloff: Duplicate key detection added in triplea-game/triplea#1759.  I also removed the duplicate `version` keys from `triplea_maps.yaml`.

RoiEXLab:
@DanVanAtta I renamed `dlurl` to `downloadUrl`
The reason I didn't just "rename" `mapName` to `title` but copied it instead is because the `title` attribute is used internally by jekyll. I didn't want any other code to rely on that.
I'm fine with the current state for now...
After this PR is merged we can of course still change stuff around without having new merge conflicts pop up.

I kept a single _maps file in order for the directory to be created. The python script does not  automatically create folders and the sh script invoking it neither. This is intentional and avoids creating unwanted folders when mistyping something.
The single file will be overidden on the first push on the triplea repo because I didn't update he dlurl var change, it will be tested as well \:-)

RoiEXLab:
@ron-murhammer @DanVanAtta could you please merge this PR (and the linked one of course) soon?
I find it quite annoying that I have to rebase this branch every time the yaml file is changed

RoiEXLab:
@DanVanAtta @ron-murhammer I resolved merge conflicts again.
I'd really appreciate it if you could merge this PR soon...

DanVanAtta:
FWIW - triplea_maps.yaml maybe would have been better deleted in a second PR

PR #192, opened 2017-05-25
Add back button

RoiEXLab:
As @DanVanAtta suggested: https://github.com/triplea-game/triplea-game.github.io/pull/185#issuecomment-303892374
This adds a back button for all the dev sites.
Current usage:
If no root value is given the root page is `../`.
If a root value is given:
  If the root value is `none` the backbutton is not created
  Else the root page is the value

This is needed, because the permalinks do not always match the actual file structure...
And 'none' is just an option for exceptions

### DIFF
--- a/_layouts/longpage.html
+++ b/_layouts/longpage.html
@@ -1,7 +1,16 @@
 ---
 layout: page
 ---
-<button onclick="topFunction()" title="Go to top" id="topScroller">Top</button>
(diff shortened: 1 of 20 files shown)


RoiEXLab:
It currently looks like this:
Screenshot
If no back button is available the top button will still be in the correct place, e.g. a couple pixels further to the right

DanVanAtta:
@RoiEXLab  testing this out - the back behavior is a bit unexpected going all the way to the home page. For more details: https://github.com/triplea-game/triplea-game.github.io/issues/193

PR #198, opened 2017-05-30
Fix footer on maps lists pages

RoiEXLab:
Before:
www triplea-game org-maps-list-categories-
After:
localhost-4000-maps-list-categories-
(It's actually worse if you have a tiny monitor)

### DIFF
--- a/_sass/_base.scss
+++ b/_sass/_base.scss
@@ -272,9 +272,11 @@ a:hover {
  * Maps Page
  */
 
-#maps-col-wrapper {
-  width: 50%;
-  float: left;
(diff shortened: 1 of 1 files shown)


_sass/_base.scss
```diff
@@ -272,9 +272,11 @@ a:hover {
  * Maps Page
  */
 
-#maps-col-wrapper {
```
RoiEXLab: This wasn't used anyway, `maps-col-wrapper` is a class not an ID

RoiEXLab:
@DanVanAtta @ron-murhammer This is a fairly simple change...

PR #202, opened 2017-05-31
Huge HTML Refactoring

RoiEXLab:
2 recommendations:
Review with `?w=1` appended to the url, lots of formatting was involved
Review one commit at a time, I made all the changes at once and created single commits to group similar changes together.
Reviewing everything but the last 3 commits should also still be pretty comprehensible. Those were made by pretty much copy-pasting and find-and-replacing stuff

No functional changes.
Every unused byte is a byte too much being sent, this should improve the loading times on slower connections.
Also this makes some html/css files easier to read.

Classes like `community_col` were changed to `community-col` in order to be consistent when it comes to dashes and underscores
src and href attributes with absolute links relative to the triplea-game.org site get and prepended baseurl if we ever wanted to change the server strucure...


ron-murhammer:
@RoiEXLab I glanced over it and looks alright. I'm not really going to review 500+ lines of html/css refactoring in detail. As long as there are no functional changes and the files are in better shape then I'm fine merging this.

RoiEXLab:
@ron-murhammer I created 3 new css files.
The mobile css file is basically just the part that used to be inside the mobile bracket now in a separate file.
Content.css contains css being used on every page/most of the pages
Misc is everything else.
If you just look at the split commit you can probably see nothing was really changed. Just the order got changed a little bit.

PR #204, opened 2017-05-31
Improve map layout

RoiEXLab:
This moves the version text to another place as suggested by @Cernelius
Looks much better now ^^
Before: http://www.triplea-game.org/map/270bc/
After:
Screenshot

### DIFF
--- a/_layouts/map.html
+++ b/_layouts/map.html
@@ -1,13 +1,13 @@
 ---
 layout: default
 ---
-<h2>{{ page.mapName }} v{{ page.version }}</h2>
+<h2>{{ page.mapName }}</h2>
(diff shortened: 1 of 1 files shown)


ron-murhammer:
@RoiEXLab Awesome work on the website. I feel like its really moving in the right direction. Now we just need to figure out a way to drive more traffic to it and the new forum!

I also added you as a committer to the website repo so we have more folks to help review PRs and you probably now know the code better than I do.

RoiEXLab:
@ron-murhammer Thanks!
I used the rights to enable travis on this repo.
I made some direct commits, because travis isn't always as straight forwards as it could be when it comes to dependencies...
For all other changes I will create a PR and self merge if something simple as a spelling mistake is fixed, as discussed on the main repo...

PR #207, opened 2017-06-02
Remove most of the content in git_setup_and_tips.md. There are other …

DanVanAtta:
…better more valuable git tutorials out there. it would be useful to have a succinct dev-workflow guide, but this not quite that, and it is not succinct.


RoiEXLab:
Merging this in order to trigger the jekyll build... (It failed because of a server outage it seems)
@DanVanAtta I believe we should add a notice stating that there are git tutorials out there and they might come in handy...

PR #214, opened 2017-06-05
Migrate 'issue labelling', copy/pasted from triplea-game wiki

DanVanAtta:
From: https://github.com/triplea-game/triplea/wiki/Issue-Labelling

Reason: we've been trying a few different ways to consolidate documentation, the latest effort was to get everything on the github.io website so it was publicly accessible. It does seem like dev documentation may move back to the triple-game/triplea code repo, but in the meantime we should complete the migration away from triplea-game wiki for any permanent documentation (wiki was decided it would be best for ephemeral lists, eg: feature requests, legacy features we are working out. Basically any type of long lived tracking list would be a candidate for wiki)

### DIFF
--- a/dev_docs/dev/issues_labels.md
+++ b/dev_docs/dev/issues_labels.md
@@ -1,0 +1,50 @@
+Labels are tags that help categorize issues. 
+Admins can create or remove them to help categorize the issue queue at any given time.
+They are pretty fluid as new categorizes of topics come up.
+Below is a listing explaining what each label should mean.
+
+
+### Bug Backlog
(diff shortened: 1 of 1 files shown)


DanVanAtta:
Opening mostly for the sake of the reason/overview. Looking at the page that is actually being migrated, it is pretty out of date! I deleted the wiki page just now, and will close this PR. I think our current label system is mostly self explanatory. The expectation is that labels will be updated/added adjusted pretty freely. With that hopefully this meta-documentation is no longer needed.

PR #215, opened 2017-06-05
Fix links

RoiEXLab:
This avoids a lot of 301 redirects, 404 and 403 errors
Also the 503 error is being supressed because the server is returning normal html for whatever reason


RoiEXLab:
I actually had a lot of fun looking up some old links via https://web.archive.org ^^
Unfortunately not all of them were present...

We really need to completely rework the legacy documentation!

dev_docs/older/readme.html
```diff
@@ -93,7 +93,7 @@
 	a game on it! Please refer to our complete game playing guide to get you
 	started on this subject.
 
-	Go to the <a href="./doc/game_playing_guide.html">Game Playing Guide</a>
+	Go to the <a href="../game_playing_guide/">Game Playing Guide</a>
```
ssoloff: Just wondering if relative links, such as the ones on this page, should be updated to the Jekyll permalinks for consistency with how we do it on other pages (e.g. `{{ '/game_playing_guide/' | prepend: site.baseurl }}`?

dev_docs/older/release_model.html
```diff
@@ -50,7 +50,7 @@
 <p>CVS will not be released as a package and placed on the main
 download site. Users can obtain a copy of the CVS anonymously through
 sourceforge. Instructions for anonymous CVS access for non-developers
-can be found <a href="http://sourceforge.net/cvs/?group_id=44492">here</a>.</p>
+can be found <a href="https://sourceforge.net/projects/triplea/">here</a>.</p>
```
ssoloff: As you said, the legacy docs need to be updated (or just removed).  The TripleA CVS repo doesn't even exist anymore. 😄

dev_docs/older/readme.html
RoiEXLab: It's not that important, because those links are relative.
If our baseurl was `/test/` e.g. the index page would be located at `triplea-game.org/test/index.html` realtive links still work because the "current directory" stays the samerelative to all the other documents.
Only when using absolute links e.g. `/downloads/` we need to append the baseurl so the link gets changed to `/test/downloads`

dev_docs/older/release_model.html
RoiEXLab: Just look at the links to maddlinks.org ^^
This appears to be an old tripleaforum which doesn't even exists anymore. It was captured a couple times on archive.org so I replaced the links with the archive links 😄

dev_docs/older/readme.html
RoiEXLab: Maybe I'm going to change that in another PR, but not now ¯\_(ツ)_/¯

PR #219, opened 2017-06-06
Add HTML5 checks

RoiEXLab:
This makes travis-based testing more strict.
HTML and css is validated, links are checked if they are reachable!


RoiEXLab:
@DanVanAtta @ron-murhammer I ran into a couple 403 Forbidden Http Codes, because it seems many travis builds are using the Github API and therefore there are no requests left.
I'd need OAuth credentials to increase the maximum limit as described [here](https://developer.github.com/v3/#increasing-the-unauthenticated-rate-limit-for-oauth-applications). They don't need write access or something. but I'd like to have this credentials being org-managed

DanVanAtta:
@RoiEXLab check the update check style script, the credential is in an environment variable. Can use that same variable?

RoiEXLab:
@DanVanAtta I guess this works :)

RoiEXLab:
@DanVanAtta I tested a couple things, and an access token works fine!
The python script works with or without one now!
I'd appreciate it if you would generate a token with minumum rights, because no rights are needed here.
Also it should be named "AUTH_TOKEN", because that's shorter ^^ (Check the python script)

ssoloff:
@RoiEXLab This is good stuff!  Just a few comments below:

1. The Travis script is a bit busy.  Would it make sense to extract portions of it to one or more shell scripts?  An additional benefit beyond readability is that a dev could more easily run these checks locally simply by invoking a shell script instead of following the "recipe" in `.travis.yml`.

1. I see that you are building some tools from source.  The danger here (more so for css-validator than your own tool) is that you are building the latest changes from the default branch (most likely `master`).  If a vendor pushes incompatible changes to `master`, it could break this build.  An improvement would be to build from a tag (or at least a specific commit) so we essentially lock down the version of the tool we're using.
&nbsp;
Even better would be to download the binary jars from a repo like Maven Central or JCenter.  The problem with css-validator, however, is the version in both of those repos does not embed the dependencies, so those jars are not runnable as-is. 😞 
&nbsp;
A third possibility is to build from source locally and commit the jars to this Git repo.  In a sense, we become our own binary repo for dependencies that are not readily available from repos like Maven Central.

RoiEXLab:
@ssoloff Yeah, I should have cleaned this up earlier ^^
I moved all of the setup stuff e.g. downloading and installing css-validator etc. to a seperate file.
I actually like using the latest master, but I agree, We should probably stick to a specific commit when installing css-validator...
I will change that!

RoiEXLab:
@DanVanAtta Once you setup the AUTH token, we should be good to go!
The build is failing once again because of the gh-api limits...

RoiEXLab:
@ssoloff I'm staying with "download the latest stable version" for the html validator, as this tool is intended as a command line tool. It's unlikely the will make breaking changes that change command line flags...
The css-validator is however not at all intended as command-line tool. It is intended as backend software for server-side languages such as PHP. (It doesn't even return non-zero exit codes when errors show up).
Therefore I agree, that changes could easily break things, especially because the css validator script is essentially a hack using grep to return non-zero exit codes...

But if you have a better suggestion, please feel free to create a PR improving the scripts after this PR is merged!

ssoloff:
Sounds good, @RoiEXLab.  As you stated, we're unlikely to get hit with this type of incompatible change.  But when one does, it tends to result in a lot of stress and fire fighting attempting to recover the "version that worked" (ask me how I know :smile:).

.travis/install_from_source
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -e
```
DanVanAtta: minor - recommend 'eux' for set flags. 'u' is just good to have, but no practical benefit here. 'x'  echo's the command being executed before executing it. I'm thinking that may make for good travis output.

.travis.yml
```diff
@@ -1,4 +1,17 @@
 language: generic
-before_install: rvm reinstall ruby-head --binary
+addons:
+  apt:
+    packages:
+      - python3
+before_install:
+  - rvm reinstall ruby-head --binary
+  - jdk_switcher use oraclejdk8
+  - ./.travis/install_from_source
```
DanVanAtta: any objection to making the script name a bit more descriptive? I'm thinking something like: 'install_html_validator' as an alternative, notably getting the name of 'what' we are installing in the script file somewhere.

.travis.yml
```diff
@@ -1,4 +1,17 @@
...
 install: gem install jekyll
-script: jekyll build
+before_script: rm ./googlecfd75a4bac031411.html
+script: 
+  - jekyll serve --detach --no-watch
```
DanVanAtta: The install is in an external script, would it make sense to extract this to an external script as well? Maybe even the same one and have flags? eg:
```
before_install:
  - ./.travis/html_validator INSTALL
:
script:
  - ./.travis/html_validator RUN
```
Please feel free to take or leave the above suggestion. I can see the suggestion as being overkill, but finding some way to make this a bit more user friendly for maintenance would be nice.

.travis/download_html_validator
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python3
```
DanVanAtta: A nice to have, not must, a quick comment in this file describing the general overview of what is happening could potentially be useful.

.travis/install_from_source
```diff
@@ -0,0 +1,14 @@
...
+mkdir ./.binary/
+cd ./.binary/
```
DanVanAtta: subshells can potentially help when doing `cd` comamnds - notably so you do not have to worry about returning back to the original directory.
EG:
```
+mkdir ./.binary/
(
 cd ./.binary/
../.travis/download_html_validator
 git clone https://github.com/w3c/css-validator.git
  (
    cd ./css-validator/
   git checkout 10409468e78e5aef25e11603c48dd8ac3b99a67d
   ant jar -q
  )
 (
  git clone https://github.com/RoiEXLab/link-checker.git
 cd ./link-checker/
 ./gradlew jar -q
  )
)
```
Not 100% sure if it makes it all that much nicer, but FWIW ^

DanVanAtta:
On second thought, let's merge this,  finish discussion here of anything - then do a second round of updates. The diff will be cleaner, and no risk of merge conflicts, and if we want to put this to bed, it'll be in a good state.

.travis.yml
RoiEXLab: The script part shouldn't be outsourced to another script, because travis treats each line as individual check.
Currently you can easily spot like: oh, check 4 failed, all the others seem ok. I must have mistyped a link somewhere...
And therefore we only need a single script. I agree, the name is probably not the best ^^

RoiEXLab:
@DanVanAtta Don't forget to add an auth-token or we are very likely to run into 403 codes in the near future

RoiEXLab:
@DanVanAtta I added a personal access token to travis...
If this token is every being leake, everyone could read all my notifications 😱 😄 
I'd really like to have this token org managed though

DanVanAtta:
Where exactly is your key being used. i'm not spotting it?

RoiEXLab:
I called it AUTH_TOKEN 
Maybe travis is not sharing those keys between different user accounts?

RoiEXLab:
Sorry, misread your answer... 
It's being used in the python script which determines the latest version of html-validator using the gh API

Issue #221, opened 2017-06-17
http://www.triplea-game.org/screenshots/ Gives 404


ron-murhammer:
http://www.triplea-game.org/screenshots/ Gives 404

The screenshot menu option gives a 404 error

ssoloff:
That page is missing a `permalink` property.  Not sure if that's the problem.  Let me spin up a Jekyll container and see if I can reproduce it locally.

ron-murhammer:
Cool. Ping me if you submit a fix.

ssoloff:
@ron-murhammer Sorry, the permalink was a red herring (it's never had one to begin with).  I'm running `master` locally and the screenshots page comes up fine.  There may be something wrong in production.

ron-murhammer:
Hmm. I'm still wondering if permalink isn't needed in production. Might be worth trying.

ron-murhammer:
@ssoloff Looks like that fixed it :). Nice work. Closing this.

ssoloff:
@RoiEXLab Can you explain why the absence of `permalink` fails in production but works fine when I run Jekyll locally?  Is there some configuration in production that's not captured in the repo?

RoiEXLab:
@ssoloff I have no idea!
I'm assuming github pages uses a slightly modified version of jekyll?

ssoloff:
@RoiEXLab Or maybe it's the Docker image I'm using that has the "extra juice" that allows it to work without a permalink?

PR #222, opened 2017-06-17
Add missing permalink (#221)

ssoloff:
Experimental fix for #221.  I'm not sure if this is required because the current version runs fine for me locally without the `permalink`.  I tested it locally to make sure the additional property didn't cause any issues.

### DIFF
--- a/screenshots.md
+++ b/screenshots.md
@@ -1,6 +1,7 @@
 ---
 layout: default
 title: Screenshots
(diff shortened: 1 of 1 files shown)


RoiEXLab:
@ssoloff The real question is why did it work locally!
Without the permalink the Screenshots should only be available via http://url/Screenshots.html

PR #226, opened 2017-06-24
Remove 'dev_docs' files, moving to 'triplea' repo.

DanVanAtta:
Couple motivations and reasons for the move:
- focuses website contents towards typical game players. This helps focus on having good help content and information useful for new and returning players
- any content that may need to be updated in sync with the code will ideally now be located in the same repo as the code. For example, if we update something relating to maps, we'll be able to then update the map maker documentation in the same branch that modifies the code.

These files are added to the game engine repo in https://github.com/triplea-game/triplea/pull/1982


RoiEXLab:
Small suggestion:
I believe the "longpage"-layout is no longer being used after this PR... It should be removed along with those files...

DanVanAtta:
I was not able to find and am curious where the longpage layout is defined. Do you know where that layout is @RoiEXLab defined?
Searching for other usages of longpage, it does look like we still have a couple:
```
game_playing_guide.html:layout: longpage
about_triplea.md:layout: longpage
```

RoiEXLab:
@DanVanAtta its a file in the layout directory

DanVanAtta:
https://github.com/triplea-game/triplea-game.github.io/issues/239 to note long page layout is likely no longer used. There may be more cleanups available. It's more important though to remove this content so it is not duplicated, and avoid the possible mistake of updating the dead copy.

PR #228, opened 2017-06-28
Add RAM documentation

RoiEXLab:
As requested in https://github.com/triplea-game/triplea/pull/2028#issuecomment-311761857

### DIFF
--- a/help/set_ram.md
+++ b/help/set_ram.md
@@ -1,0 +1,38 @@
+---
+layout: page
+title: "Set maximum RAM"
+permalink: /help/set-ram/
+---
+(Valid for TripleA v1.9.0.0.5401+)<br>
+It may happen that your TripleA client exits with an `OutOfMemoryError`, even if your PC has more RAM to offer.<br>
+In this case you might want to increase the usable amount of RAM for your installation of TripleA.<br>
+There are multiple ways to achieve this, and they all depend on the way your version of TripleA has been installed on your computer.
+
+<br>
+`<amount>` is the amount of memory you want to assign to TripleA.<br>
+`1` means 1 byte, `1K` 1024 bytes, `1M` 1024 kilobytes, `1G` 1024 megabytes and so on.<br>
+The default value TripleA uses is `1G`.<br><br>
+
+ - Installer
+   - Temporary:
+     1. Open the commandline/terminal
+     2. Enter `/path/to/triplea -J-Xmx<amount>`
+   - Permanent:
+     1. Go to your TripleA installation directory.
+     2. Open the `TripleA.vmoptions` file.
+     3. Change the `-Xmx<amount>` to an appropriate value
+     4. Start TripleA
+ - Jar file
+   - Temporary:
+     1. Open the commandline/terminal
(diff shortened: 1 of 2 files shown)


RoiEXLab:
Bumping @DanVanAtta

RoiEXLab:
@DanVanAtta awaiting your review

help/set_ram.md
```diff
@@ -0,0 +1,38 @@
+---
+layout: page
+title: "Set maximum RAM"
+permalink: /help/set-ram/
+---
+(Valid for TripleA v1.9.0.0.5401+)<br>
+It may happen that your TripleA client exits with an `OutOfMemoryError`, even if your PC has more RAM to offer.<br>
+In this case you might want to increase the usable amount of RAM for your installation of TripleA.<br>
+There are multiple ways to achieve this, and they all depend on the way your version of TripleA has been installed on your computer.
```
DanVanAtta: IMO keep it simple and provide instructions for just the latest and via installer  -  avoid giving options.

help/set_ram.md
```diff
@@ -0,0 +1,38 @@
...
+ - Installer
+   - Temporary:
+     1. Open the commandline/terminal
+     2. Enter `/path/to/triplea -J-Xmx<amount>`
+   - Permanent:
+     1. Go to your TripleA installation directory.
+     2. Open the `TripleA.vmoptions` file.
+     3. Change the `-Xmx<amount>` to an appropriate value
+     4. Start TripleA
+ - Jar file
```
DanVanAtta: I don't think we really officially support a jar file installation. We do this for bots/servers, but we really want the one path to installing for the sake of simplicity and clarity. Much easier to say, "here is how you can install", rather than "there are 3 all equally good ways to install, pick your own adventure!".

Sorry to be a bit cute, cutting options should boost clarity.

help/set_ram.md
```diff
@@ -0,0 +1,38 @@
...
+In this case you might want to increase the usable amount of RAM for your installation of TripleA.<br>
+There are multiple ways to achieve this, and they all depend on the way your version of TripleA has been installed on your computer.
+
+<br>
+`<amount>` is the amount of memory you want to assign to TripleA.<br>
+`1` means 1 byte, `1K` 1024 bytes, `1M` 1024 kilobytes, `1G` 1024 megabytes and so on.<br>
+The default value TripleA uses is `1G`.<br><br>
+
+ - Installer
+   - Temporary:
```
DanVanAtta: I'm not sure there is a use-case for a temporary update, we can probably scatch the CLI steps in favor of just having the permanent fix.

DanVanAtta:
You've done a really thorough job here, the steps are clear and cleanly layed out, easy to follow. As commented, I think you can chop off some of the 'branches' from the instructions to make them simpler. With luck should make for an even shorter and easier document to follow.  Thanks RoiEXLab for the effort and following up here!

help/set_ram.md
RoiEXLab: I was thinking about rare cases when the client isn't starting for some reason, but you are probably right

DanVanAtta:
Merging now, I think we still need to iterate on a lot of help content a couple more times to simplify it.

PR #229, opened 2017-07-12
Add preceding slash to permalinks

RoiEXLab:


RoiEXLab:
Self-merging to get rid of the 404 errors

PR #234, opened 2017-07-29
Update download page links. Reduce some of the page text, add a link …

DanVanAtta:
…to 'early releases'

- Also removed the 'upgrading from 1.8.0.9' link.

After:
updated

Before:
before

### DIFF
--- a/download.html
+++ b/download.html
@@ -27,7 +27,7 @@ permalink: /download/
 <p><a href="{{ '/release_notes/' | prepend: site.baseurl }}">Release Notes</a></p>
 <p><a id="sourceLink" href="https://github.com/triplea-game/triplea/releases/latest">Source Code</a></p>
 <p><a href="{{ '/old_downloads/' | prepend: site.baseurl }}">Previous TripleA Releases</a></p>
-<p><a id="sourceLink" href="https://github.com/triplea-game/triplea/releases/">Early Release Download</a></p>
+<p><a href="https://github.com/triplea-game/triplea/releases/">Early Release Download</a></p>
 
 
 <div id="windows-install" class="installation-instructions">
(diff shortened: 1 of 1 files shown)


download.html
```diff
@@ -23,10 +23,12 @@
...
 
-<p>Upgrading from 1.8.0.9? <a href="{{ '/upgrade_notes/' | prepend: site.baseurl }}">read here for upgrade notes</a></p>
-<p>See what has changed in the <a href="{{ '/release_notes/' | prepend: site.baseurl }}">release notes</a></p>
-<p>Download <a href="{{ '/old_downloads/' | prepend: site.baseurl }}">previous TripleA releases here</a></p>
-<p>Download <a id="sourceLink" href="https://github.com/triplea-game/triplea/releases/latest">Source Code from Github Releases</a></p>
+<h2>Additional Links</h2>
+<p><a href="{{ '/release_notes/' | prepend: site.baseurl }}">Release Notes</a></p>
+<p><a id="sourceLink" href="https://github.com/triplea-game/triplea/releases/latest">Source Code</a></p>
+<p><a href="{{ '/old_downloads/' | prepend: site.baseurl }}">Previous TripleA Releases</a></p>
+<p><a id="sourceLink" href="https://github.com/triplea-game/triplea/releases/">Early Release Download</a></p>
```
RoiEXLab: You used some duplicate ids here which caused the travis build to fail.
You should probably use classes instead

DanVanAtta: hmm, i wonder then if we really need the ids?

RoiEXLab: I don't know, you added them :D

DanVanAtta: Was following the existing pattern a bit blindly. Removed the extra id, it turns out it is used in javascript. The fact we have a value for the link is a bit misleading, it is overwritten.

PR #235, opened 2017-08-18
Correct false statement that GPL is non-commercial

trevorld:
Not only does the GPL allow commercial use but in fact commercial use is one of the 4 basic freedoms that the GPL seeks to preserve (and forbidding commercial use means you can't be hosted in the Debian Linux repositories among other free software repositories).  Commercial distribution of GPL code is how Sourceforge makes money (i.e. showing ads).  However the fact that you must share the source code and allow others to freely give away your changes does make it hard to make money commercially from a Triple A fork...

https://www.gnu.org/licenses/gpl-faq.html#GPLCommercially

### DIFF
--- a/about.md
+++ b/about.md
@@ -7,7 +7,7 @@ title: About
 TripleA is a networked open source strategy gaming engine that allows for customized user editable maps and rules. Typically game rules are based upon Axis and Allies rules.
 
 
-TripleA is released under the GNU Public License (GPL). This license means the work is completely non-commercial, and is not allowed for use as part of other commercial projects.
+TripleA is released under the GNU Public License (GPL). This license means you are free to distribute this work and any derivatives you make from this work either commercially or non-commercially but only under the terms of the GPL (i.e. must make source code available to your users and must allow your users themselves to freely distribute the work and any further derivatives under the GPL).
 
 ## About TripleA
 


ssoloff:
Agree with this.  I suspect there was some confusion between the terms _commercial_ and _proprietary_, the former being permitted as described by the OP, while the latter is explicitly prohibited.

However, I would suggest that we simply use some FSF boilerplate for this license instead of getting into the weeds about commercial usage.  Also, we probably want to update the text to make it clear that we're using GPL v2.  I'll offer up the standard GPL notice for consideration:

> TripleA is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; either version 2 of the License, or (at your option) any later version. TripleA is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the [GNU General Public License](https://www.gnu.org/licenses/old-licenses/gpl-2.0.html) for more details.

trevorld:
I'm not sure if Triple A is GPL-2 or GPL-2+ but if it is the former the standard boilerplate notice should be:

This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; **version 2.**  ~~either version 2 of the License, or (at your option) any later version.~~

ssoloff:
FWIW, the original license text Sean added to the source in 2002 included the "any later version" clause (see [triplea-game/triplea@6b5a957](https://github.com/triplea-game/triplea/commit/6b5a957#diff-b4627b82acbc76c03e10ccd117567e6b)).  @ron-murhammer, @DanVanAtta, or @veqryn may know if this licensing option was ever removed.

veqryn:
👍

DanVanAtta:
This is a good correction, thank you!

PR #240, opened 2017-08-25
About.md edits

DanVanAtta:
- give licensing its own section and drop to bottom. The location is more in consideration of the other topics we have in the page, they seem to belong together before licensing info.
- clean up first few paragraphs a bit, shorten the text, reword
- update heads, drop the sub-section for TripleA features, the subsections can be their own h2 headers. This is to simplify the structure of the page a bit more.

Still a bit more to be done for future iterations:
-  the newly labelled history section does not seem very complete, the history info ends with a future sounding note, but from a perspective that ends in around 2005.
- We also still have similar info on other pages, some de-duplication would be worthwhile, it's also not clear if all of the content on 'about.md' should live there.

### DIFF
--- a/about.md
+++ b/about.md
@@ -4,26 +4,23 @@ permalink: /about/
 title: About
 ---
 
-TripleA is a networked open source strategy gaming engine that allows for customized user editable maps and rules. Typically game rules are based upon Axis and Allies rules.
+
+TripleA is a networked open source strategy gaming engine that allows for Axis and Allies style rules and customized user editable maps.
 
 
-TripleA is released under the GNU Public License (GPL). This license means you are free to distribute this work and any derivatives you make from this work either commercially or non-commercially but only under the terms of the GPL (i.e. must make source code available to your users and must allow your users themselves to freely distribute the work and any further derivatives under the GPL).
+## TripleA History
 
-## About TripleA
-
-TripleA was started in Oct 2001 by an unemployed Sean Bridges who was looking for a way to build up his resume after the end of the tech bubble. Unfortunately Sean was too successful, and soon after the 0.1 release (which he managed to demo in a job interview) he secured gainful employment.
+TripleA was started in Oct 2001 by an unemployed Sean Bridges looking for a way to build up his resume after the end of the tech bubble. Using the 0.1 release as a demo in a job interview, Sean was soon employed.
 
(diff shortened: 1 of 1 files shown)


about.md
```diff
@@ -58,7 +55,12 @@ Though the majority of our maps and games revolve around World War II, we also h
...
-### User Created Content:
+## User Created Content:
 * TripleA contains the ability to create and download user made maps.
 * Many of these maps, like "New World Order" have generated quite a following.
 * Rules can vary by map, depending on what the map creator wants.
+
+## Software Licensing
+
+TripleA is released under the GNU Public License (GPL). This license means you are free to distribute this work and any derivatives you make from this work either commercially or non-commercially but only under the terms of the GPL (i.e. must make source code available to your users and must allow your users themselves to freely distribute the work and any further derivatives under the GPL).
+
```
ssoloff: Related to #235, it would probably be a good idea to state the GPL version in this paragraph, as well as providing a link to the full license text.  This can all be jammed into the first sentence:

> TripleA is released under the [GNU General Public License](https://www.gnu.org/licenses/old-licenses/gpl-2.0.html) (GPL) version 2 or later.

ssoloff:
Just to clarify...  I wasn't suggesting that _all_ of the text after the first sentence be removed.  I was only requesting a change to the first sentence.

However, I _did_ suggest that this entire paragraph be replaced with GPL boilerplate back in [#235](https://github.com/triplea-game/triplea-game.github.io/pull/235#issuecomment-323528916), but I'm ok with the second and subsequent sentences as-is.

DanVanAtta:
@ssoloff are you okay opening a quick follow PR with your suggestions? If so please feel free to merge this and continue, otherwise please let me know if you think it is best that those changes be included with this branch before we take another iteration.

DanVanAtta:
For my motivations, I'm most concerned the GPL text we have is accurate, but beyond that I'm a bit indifferent. I do think a person reading the 'about' page is most likely to be interested in the game history and features instead of the licensing details. I could be wrong, but the theme of the site has been to optimize for new potential players and get them to join and play the game. While we still cater to someone digging up license details by including the content, I think we primarily should be focused on new potential players first and utmost (then existing players, and then any other audience)

ssoloff:
@DanVanAtta NP, I'll submit a follow-up PR.

PR #244, opened 2017-08-26
Improve overall generation

RoiEXLab:
This PR is a collection of small improvements:
1. The rss page no longer generates non-existing post pages.
2. We make use of jekyll 3.5.2/liquid 4.0.0, and clear out uneccessary whitespaces in page generation (inside of loops for example)

There are no functional changes!

### DIFF
--- a/_layouts/default.html
+++ b/_layouts/default.html
@@ -1,15 +1,15 @@
 <!DOCTYPE html>
 <html lang="en">
-  {% include head.html %}
+  {% include head.html -%}
   <body>
     <div id="container">
-      {% include header.html %}
+      {% include header.html -%}
       <div id="body">
         <div id="page-content">
           {{ content }}
         </div>
       </div>
-      {% include footer.html %}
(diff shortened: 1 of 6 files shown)


_layouts/default.html
```diff
@@ -1,15 +1,15 @@
 <!DOCTYPE html>
 <html lang="en">
-  {% include head.html %}
+  {% include head.html -%}
```
ssoloff: Can you please explain what the hyphens after/before the open/close tags do?  I can't find it in the Jekyll docs.  Are these related to the whitespace collapse you mentioned in the PR description?

RoiEXLab: You can find them in the liquid docs, which is the preprocessor jekyll uses: https://shopify.github.io/liquid/basics/whitespace/
They basically remove whitespace before and after the tags and were introduced in liquid 4.0.0 which is used since jekyll 3.5.0

PR #248, opened 2017-08-27
Improve bug reporting page

### LINKED ISSUE #17: Bug Report Page - Clear and easy to access bug reporting instructions
There is enough information and detail around bug reports that it could use it's own page. Maybe under website top level "help" category. IE:   http://triplea.../helps/ReportingBugs.html

Page would contain the details of what to include in the report, and then some links and maybe screenshots of actually how to submit the report.

I'm curious if others agree there is enough detail around bug reports that it could use it's own treatment. I could even see reason to have a "report a bug" link in the header links.

RoiEXLab:
Goal of this PR is to resolve #17
@DanVanAtta suggestions are welcome, did I miss something, could something be improved?

### DIFF
--- a/help/bug_report.md
+++ b/help/bug_report.md
@@ -4,13 +4,20 @@ title: Bug Reports
 permalink: /bug_report/
 ---
 
-* <a href="https://github.com/triplea-game/triplea/issues?q=is%3Aissue+is%3Aopen+label%3ABug">Open Bug List</a>
+* [Open Bug List](https://github.com/triplea-game/triplea/issues?q=is%3Aissue+is%3Aopen+label%3A%22type%3A%20bug%22)
 
-## Filing a bug report:
+## Before doing anything
 
-* Create a <a href="https://github.com/account"> github account</a>
-* Submit a tracking issue: <a href="https://github.com/triplea-game/triplea/issues/new">https://github.com/triplea-game/triplea/issues/new</a>, you'll need to give the topic a title, and fill out a description of what you would like done.  Please try to be as concise and specific as you can. Always please try to give specific examples of problems, map names, etc, to help make it very easy to focus in directly on what the problem is. 
+Before submitting a bug report at all, take a step back and make sure it's really a bug and not a feature request.
+Feature Request often need to be discussed first, if you still want to submit your feature Request, please do so in [our Forum](https://forums.triplea-game.org/category/42/feature-requests-ideas).
+If you are not sure if you are experiencing an actual bug, or it's just your own System which makes TripleA not work properly for you, [our community will be willingly to help](https://forums.triplea-game.org/category/10/help-questions).
+
+## Filing a bug report
+
+* Make sure that your bug wasn't already reported.
+* Create a [GitHub Account](https://github.com/account)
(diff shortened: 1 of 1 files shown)


help/bug_report.md
```diff
@@ -4,13 +4,20 @@ title: Bug Reports
...
 
-* <a href="https://github.com/triplea-game/triplea/issues?q=is%3Aissue+is%3Aopen+label%3ABug">Open Bug List</a>
+* [Open Bug List](https://github.com/triplea-game/triplea/issues?q=is%3Aissue+is%3Aopen+label%3A%22type%3A%20bug%22)
 
-## Filing a bug report:
+## Before doing anything
 
-* Create a <a href="https://github.com/account"> github account</a>
-* Submit a tracking issue: <a href="https://github.com/triplea-game/triplea/issues/new">https://github.com/triplea-game/triplea/issues/new</a>, you'll need to give the topic a title, and fill out a description of what you would like done.  Please try to be as concise and specific as you can. Always please try to give specific examples of problems, map names, etc, to help make it very easy to focus in directly on what the problem is. 
+Before submitting a bug report at all, take a step back and make sure it's really a bug and not a feature request.
```
DanVanAtta: I recommend striking this first sentence. We would need to tell our users how to determine if something is existing or not, and it is all extra hurdles for something that needs to be easy (like stupid easy). The natural reaction of most users at just seeing this page and the amount of text is probably to close their browser window, let alone after seeing how much stuff they have to do.

I also think that de-duping bug reports is our jobs as admins, our bugs better not be so pervasive, otherwise we have a p0, or we have a pretty important update to make since we are getting so much feedback.

help/bug_report.md
```diff
@@ -4,13 +4,20 @@ title: Bug Reports
...
+Feature Request often need to be discussed first, if you still want to submit your feature Request, please do so in [our Forum](https://forums.triplea-game.org/category/42/feature-requests-ideas).
```
DanVanAtta: Recommend striking this next sentence, instead create a section to capture this content, something like:

```
"## Send us Suggestions"

* What would you like to see improved? You can let us know here *link*
```

The reason I suggestion this is for skimming value, assume someone just had a game crash, they lost 5 hours of play and an interesting game, we are lucky almost at all if they will give us one more click. By that token, we need directions that are really dead simple, here is how you submit a bug report, one two three.

help/bug_report.md
```diff
@@ -4,13 +4,20 @@ title: Bug Reports
...
+If you are not sure if you are experiencing an actual bug, or it's just your own System which makes TripleA not work properly for you, [our community will be willingly to help](https://forums.triplea-game.org/category/10/help-questions).
```
DanVanAtta: Again, I would get right to the bug report instructions, this link may be good in a new  'FAQ' section, or a 'More Help' section that would be found further down.

help/bug_report.md
```diff
@@ -4,13 +4,20 @@ title: Bug Reports
...
+
+## Filing a bug report
```
DanVanAtta: Recommend re-wording, maybe "Having a Problem? Let us know!". Or "Send us a Bug Report". Or "How to submit a Bug Report".

"Filing" is really formal, something you do at a government agency before they ignore and then lose your paperwork ;)

DanVanAtta:
https://github.com/triplea-game/triplea-game.github.io/pull/250 has a few updates of what I'm thinking @RoiEXLab more concretely . Please have a look and let me know what you think.

RoiEXLab:
Liking your changes, closing in favour of #250

PR #251, opened 2017-08-29
Use correct urlescape filter

RoiEXLab:


_layouts/map.html
```diff
@@ -7,6 +7,6 @@
 <img src="{% if page.img %}{{ page.img }}{% else %}{{ '/images/missing_map.png' | prepend: site.baseurl }}{% endif %}" alt="Map Thumbnail">
 {{ content }}
 <br>
-<a href="triplea:{{ page.mapName | escape | replace: " ", "%20"  }}" class="button">Download with TripleA</a>
+<a href="triplea:{{ page.mapName | escape | url_escape }}" class="button">Download with TripleA</a>
```
RoiEXLab: keeping the initial escape, to prevent quotes in the url from breaking the html

ssoloff:
@RoiEXLab Looks like the build is failing because, for some reason, the spaces are not being escaped.  Can you please take a look to see if there's some kind of interaction between `escape` and `url_escape` that is preventing `url_escape` from working as expected?

RoiEXLab:
@ssoloff I'm currently testing... It seems the url_escape as well as the url_param_escape filters are completely broken. The do not work at all :(

Wanted to close this, but found the issue :D 
It seems this filter was renamed to url_encode accourding to the [liquid source](https://github.com/Shopify/liquid/blob/master/lib/liquid/standardfilters.rb#L44), this works, but produces + instead of %20, but this seems to not matter as we already noticed :)

ssoloff:
@RoiEXLab [This PR](https://github.com/Shopify/liquid/pull/259) seems to suggest that `url_escape` and `url_param_escape` are not part of the core liquid project (and a search of the code seems to confirm this).  I'm not sure why they're then [documented](https://help.shopify.com/themes/liquid/filters/string-filters) in, what appears to be, the "standard" liquid filters.  (I was surprised that no error was raised about a missing filter, but this appears to be the [default behavior](https://github.com/Shopify/liquid#undefined-variables-and-filters) for liquid.)

According to the [Jekyll docs](https://jekyllrb.com/docs/templates/), in addition to the standard liquid filters, Jekyll supports `uri_escape`, which appears to have the same behavior as `url_escape`.

As discussed in triplea-game/triplea#2290, RFC 3986 states that `%20` is the correct way to escape a space in the _path_ portion of a URI (which the map name effectively is for a `triplea:` URI).  See the end of this [SO answer](https://stackoverflow.com/a/29948396/3900879) for a succinct statement about how to encode a space within different portions of a URI.

Therefore, I would suggest trying `uri_escape`, but, if that doesn't work or is not available, then just close the PR, as what you had originally is correct from a standards perspective vs. what `url_encode` is doing (even though it would have been nice to avoid manually escaping spaces with `%20`).

RoiEXLab:
@ssoloff Yup, works updated the PR accordingly

PR #253, opened 2017-10-15
Add partial release notes for build 7062

ssoloff:
Addressing @DanVanAtta's [comment](https://github.com/triplea-game/triplea/issues/1768#issuecomment-336670272) that we need to work on the release notes for 7062.

I added the PRs I authored that I felt would have significance to an end-user.  Some of those have questionable value, so please feel free to suggest their removal.  Note also that some of those probably made it into the 5802 and 5855 releases, but since those are no longer considered official releases, I just lumped them all under 7062.

@DanVanAtta @RoiEXLab @ron-murhammer @simon33-2 Not sure if everybody wants to just push commits to this branch with their contributions, or if we should  do it in separate PRs.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -12,6 +12,11 @@ permalink: /release_notes/
 * Improve security of network game authentication (#2107).
 * Add option to securely save PBEM/PBF credentials (#1955).
 * Fix out-of-date map check (#1695).
+* Fight a single battle automatically
+* Upgrade retreat options - no double prompt with one valid territory
+    - don't offer prompt when only defenseless units remain
+- Show dice details by default in PBF/PBEM
+- Auto fight battles with no defenders who can fire and no dependent battles
 
 ## 1.9.0.0.5855 - July 22nd 2017
 
@@ -38,8 +43,6 @@ permalink: /release_notes/
 
 * Primarily bug fixes. SBR rules now working again on most maps (map XMLs updated, download of latest maps required)
 
-
-## Changes for 1.9.0.0
 
 * Minimum Java version is now Java 1.8
 * New Install4j Installers available for all OS, replaces makensis installers
--- a/release_notes.md
+++ b/release_notes.md
@@ -14,9 +14,24 @@ permalink: /release_notes/
 * Fix out-of-date map check (#1695).
 * Fight a single battle automatically
 * Upgrade retreat options - no double prompt with one valid territory
-    - don't offer prompt when only defenseless units remain
-- Show dice details by default in PBF/PBEM
-- Auto fight battles with no defenders who can fire and no dependent battles
+  * don't offer prompt when only defenseless units remain
+* Show dice details by default in PBF/PBEM
+* Auto fight battles with no defenders who can fire and no dependent battles
+* Add experimental savegame format (#2128)
+* Setting Menu Overhaul (#2137)
+* Bugfix: Maps can have AI-only players(#2149)
+* Removal of the Beta Map Creator (#2156)
+* Enabled TripleA to download Maps via a browser (#2274)
+* Updated many outdated links
+* Improved internal password security (#2101)
+* General smaller performance improvements
+* Several smaller usability improvements
+* Using Java 1.8.0_144 for the installer now (#2384)
+* Removed the 'old jar' functionality (#2284)
+
+And many more internal changes, bugfixes etc.
+See the full List [here](https://github.com/triplea-game/triplea/pulls?utf8=%E2%9C%93&q=merged%3A%3E2017-07-22)
+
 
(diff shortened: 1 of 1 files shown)


simon33-2:
Added some stuff that I'm aware of to a pr in your branch.

release_notes.md
```diff
@@ -30,8 +60,6 @@ permalink: /release_notes/
 * Primarily bug fixes. SBR rules now working again on most maps (map XMLs updated, download of latest maps required)
 
 
-## Changes for 1.9.0.0
```
RoiEXLab: Why is this title being removed?

RoiEXLab:
@ssoloff I added a commit with a couple more release notes.
Once we have all release notes, I'd suggest to sort them in alphabetical or chronological (or PR number based) order and to have a consistent wording scheme:
E.g.
 - Added feature xyz
 - Fixed bug zxy
 - Redid zyx

instead of 
 - Add more of abc
 - Fixed a bug where def
 - Changes to cde

release_notes.md
```diff
@@ -4,6 +4,35 @@ title: Release Notes
...
+* Enabled TripleA to download Maps via a browser (#2274)
+* Updated many outdated links
+* Improved internal password security (#2101)
+* General smaller performance improvements
+* Several smaller usability improvements
+* Using Java 1.8.0_144 for the installer now (#2384)
+* Removed the 'old jar' functionality (#2284)
+
+And many more internal changes, bugfixes etc.
+See the full List [here](https://github.com/triplea-game/triplea/pulls?utf8=%E2%9C%93&q=merged%3A%3E2017-07-22)
```
RoiEXLab: @ssoloff btw. I used this link to get all PRs being merged since the last entry in this file, I didn't paricularily checked who created Pr xyz, I just looked at the title and copied the PR number into the changelog

release_notes.md
ssoloff: Bumping for @simon33-2.

release_notes.md
ssoloff: Thanks, @RoiEXLab.  And I agree about having some kind of consistent naming convention to differentiate features, bug fixes, and "everything else".  As well as choosing a consistent sort order. :smile:

release_notes.md
```diff
@@ -13,6 +42,7 @@ permalink: /release_notes/
 * Improved naming of auto save games.
 * Set all buttons to select alliances more easily
 * Bug Fix: Fix game crash when game release version is updated
+* Associate TripleA savegames with TripleA
 
 ## 1.9.0.0.3627
```
ssoloff: I'm wondering if we should update this to `3635` to be consistent with the build that was _actually_ marked as stable.

release_notes.md
```diff
@@ -4,6 +4,35 @@ title: Release Notes
...
+* Fix threading errors reported when starting Map Creator (#2238).
+* Improve security of network game authentication (#2107).
+* Add option to securely save PBEM/PBF credentials (#1955).
+* Fix out-of-date map check (#1695).
+* Fight a single battle automatically
+* Upgrade retreat options - no double prompt with one valid territory
+  * don't offer prompt when only defenseless units remain
+* Show dice details by default in PBF/PBEM
+* Auto fight battles with no defenders who can fire and no dependent battles
+* Add experimental savegame format (#2128)
```
ssoloff: I'm going to suggest removing this line.  Since it's an optional beta feature, disabled by default, and not yet fully implemented, I don't think it will be meaningful to the average user.

release_notes.md
ssoloff: Another possibility is to use something like the format suggested by [keep a changelog](http://keepachangelog.com).  That is, instead of identifying items as _features_, _bug fixes_, etc., we group them by _added_, _changed_, and _removed_.

release_notes.md
DanVanAtta: sounds good

release_notes.md
DanVanAtta: ok

DanVanAtta:
This is a good working list. I think squash and merging now is best so that we can keep submitting PRs for updates. A PR into this branch would be against a fork repo, and AFAIK would show up under the PR tab of that fork, not in this org.

@ssoloff I think we may do well with a branch that is pushed to the triplea-game.github.io repo, then we can submit PRs into that branch and potentially accumulate more changes. Perhaps more useful for next time we have a new latest release.

Note: We get about 1-5 unique visitors per day, likely not even visiting this page. It's better trade off IMO to keep this lightweight and allow anyone to rebase and offer more suggestions

release_notes.md
simon33-2: Looks like a meaningless line. Perhaps a typo but I don't think it matters much either way.

PR #256, opened 2017-10-18
Use consistent grammar and punctuation in release notes

ssoloff:
Some cosmetic changes so it doesn't read like four different people wrote the release notes: :smile:

* Capitalize first word on each line.
* Remove trailing periods.
* Use active voice.

I also corrected the date of the 7062 release (somehow I flubbed it by three days when I submitted the original PR).  I added an upper bound to the PR search URL so it doesn't return PRs merged after 7062 was released and moved the lower bound to the 3635 release date.

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -7,37 +7,37 @@ permalink: /release_notes/
 ## 1.9.0.0.7062 - October 14th 2017
 
 ### Added
-* Add 'remember me' option when logging in to lobby, credentials are to securely saved on your computer (#2395).
-* Add option to securely save PBEM/PBF credentials (#1955).
+* Add 'remember me' option when logging in to lobby, credentials are securely saved on your computer (#2395)
+* Add option to securely save PBEM/PBF credentials (#1955)
 * Set all buttons to select alliances more easily in single player mode
 
 ### Changed
 * Upgrade retreat options - no double prompt with one valid territory
-  * don't offer prompt when only defenseless units remain
+  * Don't offer prompt when only defenseless units remain
 * Game play speedup, automatically select a single battle for combat
-* Game play speedup, automatically select defenseless unit battles first. For example, when selecting battles to fight, all battles against undefended transports would be auto-selected first.
-* Setting Window available from the file menu was overhauled (#2137)
+* Game play speedup, automatically select defenseless unit battles first (e.g. when selecting battles to fight, all battles against undefended transports would be auto-selected first)
+* Overhaul Settings Window available from the file menu (#2137)
 * Show dice details is now a default option in PBF/PBEM
-* Enabled TripleA to download Maps via a browser (#2274)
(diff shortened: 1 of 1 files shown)


release_notes.md
```diff
@@ -4,43 +4,42 @@ title: Release Notes
...
-* console opening when playing a network game (#2289).
-* threading errors reported when starting Map Creator (#2238).
-* out-of-date map check (#1695).
-* Maps can have AI-only players(#2149)
-* Updated many outdated game links
+* Console opening when playing a network game (#2289)
+* Threading errors reported when starting Map Creator (#2238)
+* Out-of-date map check (#1695)
+* Maps can have AI-only players (#2149)
+* Update many outdated game links
```
RoiEXLab: This doesn't sound very good if you read through all the changelog.

release_notes.md
```diff
@@ -4,43 +4,42 @@ title: Release Notes
...
 * Show dice details is now a default option in PBF/PBEM
-* Enabled TripleA to download Maps via a browser (#2274)
-* Associate TripleA savegames with TripleA at the OS level when running the installer.
-* Improved naming of auto save games.
-* Improved internal password security (#2101)
-* Improve security of network game authentication (#2107).
+* Enable TripleA to download Maps via a browser (#2274)
+* Associate TripleA savegames with TripleA at the OS level when running the installer
+* Improve naming of auto save games
+* Improve internal password security (#2101)
```
RoiEXLab: The heading is changed, while this is written in simple present, this doesn't sound that good either :/

ssoloff: I have no problem changing the text to use a passive voice, as long as everything is consistent. :smile:

However, note that

1. most Git best practice guides recommend using [imperative commit messages](https://chris.beams.io/posts/git-commit/#imperative), and
1. the proliferation of tools to automatically generate changelogs from Git commit messages (e.g. [here](https://github.com/semantic-release/semantic-release))

means that changelogs will generally be written in an active voice.

I guess a different question is should we write the text assuming the heading (e.g. Added, Fixed, etc.) is the first word of each sentence?  Even the [keepachangelog](http://keepachangelog.com/en/1.0.0/) changelog example inconsistently switches between a) assuming the heading is the first word and b) active voice that disregards the heading. 😕

RoiEXLab: :/
Nevermind then.
As long as we all know what we meant, we should be alright

DanVanAtta: My vote is for passive voice. eg: "We have 'released' a new version, here is the work we _did_"
Our audience is game players, not developers. Most commit messages are not going to be that useful to them and should be trimmed down to a useful summary of what changed.

DanVanAtta: Hmm, not the best phrased, some commit / merges we can omit from the notes, otherwise we likely could use more detail about the implications of these changes to make the notes more useful for our game players.

PR #258, opened 2017-10-20
Rollback release number in release notes

DanVanAtta:
Download map windows is unusable: https://github.com/triplea-game/triplea/issues/2511, given that bug we rolled back to 3635 as the latest release.


ssoloff:
@DanVanAtta triplea-game/triplea#2511 was introduced in build 7137.  It is not present in build 7062 (confirmed on my Windows install).

PR #263, opened 2017-11-04
Add missing PR number

ssoloff:


ssoloff:
Self-merging as this is basically just a typo fix.

PR #268, opened 2017-11-25
Fix community page mobile layout

RoiEXLab:
This PR fixes the community page on mobile layouts and improves usability in general:
Changes:
 - The layout now uses a flexbox instead of floating divs, which don't have any real width for the css parser (that's what was causing the fotter to be over the site)
 - Replaced the icons with up to date icons (Github uses an svg, a&a.org also has an updated logo)
 - Adjusted the icons to all have the exact same size, so the text under it has the same height
 - The logos are now clickable just like the titles above them

Examples:
Before:
Desktop Before
Mobile Before
After:
Desktop After
Mobile After
(The last screenshot is just cut-off the page looks normal)


RoiEXLab:
Tagging @ron-murhammer for this then to have a second look

PR #270, opened 2017-11-26
Update release notes after 7527 was unreleased.

DanVanAtta:


release_notes.md
```diff
@@ -8,8 +8,6 @@ permalink: /release_notes/
 
 See the [full list here](https://github.com/triplea-game/triplea/pulls?q=merged%3A%3E%3D2017-11-14T00%3A21%3A00-05%3A00).
```
ssoloff: You'll need to delete this line and replace the link in the next "See the full list here" on line 24 with 

```
https://github.com/triplea-game/triplea/pulls?q=merged%3A%3E%3D2017-11-02T18%3A11%3A00-04%3A00
```

DanVanAtta: Can we reconsider the the 'see the full list' link? We are very close to having releases be just a check of a box. Any way we can make it so this link is not an overhead item to update?

DanVanAtta: The audience of the change list are game players, we should not expect them to jump into PRs or commit comments. Let's clarify the audience, be sure we are on the same page there, and then suggest alternatives if not outright deleting the link or replacing it with something that never needs to be updated.

ssoloff: Well, the whole _release\_notes.md_ file is kind of a blocker for having releases be just a check of a box because we have to manually update it. :smile:  If generating the link manually is seen as slowing down the process, it shouldn't be too difficult to write a script that uses the GitHub API to automatically generate it given one or two release tags (the former being for the Unreleased link).

> The audience of the change list are game players, we should not expect them to jump into PRs or commit comments.

I believe that's why we buried the merged PRs behind a link so they wouldn't be in the face of a casual release notes reader.  I think @RoiEXLab's original motivation for the link was to provide a complete list of changes that went into a release without having to document everything in markdown.  Personally, I've found these links useful in two situations:

1. When I've had to update _release\_notes.md_ with the actual markdown changes, especially when it comes to PRs I didn't author and may not be in the front of my mind for inclusion in the release notes.
1. When triaging bug reports for regression status.  This has only become useful to me lately due to the large number of releases we've had in the past month or so compared with earlier in the year.

DanVanAtta: I don't think I have a problem expanding the audience of the release notes to include developers. 

Release note updates are missing from our release steps: https://github.com/triplea-game/triplea/blob/master/docs/admin/how_to_release.md. We have not really hammered out who/when updates the notes, grant-it, the current ad-hoc process has so far worked.

Maintaining this link is not a problem, would be ideal to be sure to document how it is generated at least. I think of most importance is for us to have a mostly complete list of release steps. It's a good basis for further process improvements, and it is happily getting close to just checking a box.

ssoloff:
Closing, as this PR has been superseded by #271 with the release of 1.9.0.0.7621.

@DanVanAtta I added tasks to my TODO list to add a small script to this repo to generate the release note links and to update the How to Release document, as needed.  Hopefully will get to that soon.

PR #272, opened 2017-11-30
Merge "de-released" changelog with actual release

RoiEXLab:
Basically #270 on top of  #271
@ssoloff Please tell me what you think, I'm not sure if it's worth it but just for consistency...

### DIFF
--- a/release_notes.md
+++ b/release_notes.md
@@ -10,16 +10,6 @@ See the [full list here](https://github.com/triplea-game/triplea/pulls?q=merged%
 
 ## [1.9.0.0.7621](https://github.com/triplea-game/triplea/releases/tag/1.9.0.0.7621) - November 25th 2017
 
-### Fixed
-* Air units not able to attack after being transported by allied carrier (#2634)
-* Unexpected hits to multi-hit units in Hard AI games (#2617)
-* Slow battle calculator in network games (#2617)
-* Bot crash when joining game (#2610)
-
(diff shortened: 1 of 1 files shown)


ssoloff:
@RoiEXLab Nothing wrong with this.  I was actually thinking about doing it earlier when I closed #270.  Then I wondered if it would be better to not "rewrite history" and simply mark 7527 as withdrawn/deprecated/whatever in the release notes.  I believe this is the only log we have of what releases were actually promoted to stable status.

It _was_ released and was out there for over a week, so there might have been a significant number of users who installed it.  And, if you remember, @ron-murhammer was ready to promote 7621 to stable on 11/25 before 7527 was de-released that afternoon.  If we would have promoted 7621 before 7527 was de-released, then 7527 would still be considered an official stable release today.  Seems kind of weird.

OTOH, we probably don't want users to look back through the release notes and say, "7527 was a stable release, so I should be able to install it even if it's not the latest."  Because of that, we definitely need to change something.  Either what's proposed in this PR or, as I mentioned above, marking up 7527 in such a way that user's know not to install it.

I'm going to approve this, but it would be nice if @ron-murhammer or @DanVanAtta could weigh in with their opinion.

ron-murhammer:
@RoiEXLab @ssoloff This looks fine to me. I think having it cleaner is probably better given 7527 wasn't out very long and never promoted in the lobby or forum.

PR #286, opened 2018-01-23
update readme with ubuntu jekyll latest install steps.

DanVanAtta:


README.md
```diff
@@ -18,6 +18,13 @@ If you are interested in making changes to the design, functionality, or structu
 
 `gem install jekyll`
 
+
+Ubuntu:
+```bash
+sudo apt-get install ruby ruby-dev make gcc
```
RoiEXLab: I'd recommend using `apt` instead of `apt-get`, but not that important

PR #287, opened 2018-01-23
Remove about page

DanVanAtta:
- History content from that page has been moved to forum post: https://forums.triplea-game.org/topic/523/an-early-triplea-history-from-svn-sourcecode-webpage
- License text moved to bottom of main page
- Rest of the content should be really self explanatory and evident by checking out the screenshots, the 'live multiplayer and single player' game mode text on index.html page.

Perhaps of controversay a good bit of the about/features content is to describe and differentiate TripleA from other game clones, with some really notable features. At the same time the feature list sounds like a resume, "includes sound". So it's a bit tough to rewrite this content and boil it down, and eventually we have most of it or it should be pretty self-evident from the rest of the website.

With that there is nothing left on the about page, I was planning to rename it 'features' once the history content was removed and license moved, but makes sense to prune it altogether after those two updates.


_includes/footer.html
```diff
@@ -3,8 +3,6 @@
     <div class="footer-col">
       <h3>About</h3>
       <ul>
-        <li><a href="{{ '/about/' | prepend: site.baseurl }}">About TripleA</a></li>
```
ssoloff: Should there still be a "History" link to the new forum post?  It seems it will be difficult to find, given that it's not a stickied post.  The history was the best part of the About page. :smile:

DanVanAtta: At some point maybe, I think we need to evolve the history and make an updated thread for it. Agree history was best part, but it's about 13~15 years dated at this point. I think it was most interesting for developers, on website it's a tough call to prune it so we can focus on main mission of presenting the game and getting people to download it. The footer links I agree would be a good place for it, but again, in current form it's weak content : |

DanVanAtta: Important to mention, website was intended to be a repo of static content and help information, to remove that content from forum posts. This does not seem to have worked out very well, so we are moving back the other way. Forums is where the information is best kept right now. Given that, we also don't want website to just be a table of links to the forum content. So I'm trying to refocus website here to have a cleaner mission of presenting the game to new users for download. We do have the forums link in header, so if someone does goes spelunking in forums they'll find the content easily enough I hope.

index.html
```diff
@@ -15,6 +15,14 @@ <h2>The World's Foremost Open Source Grand Strategy Game</h2>
...
       <p>100% open source, community run, free to play game</p>
-      <p>Multiplayer and single player game options</p>
+      <p>Play live online via lobby, play by email or versus AI</p>
     <a href="{{ '/maps/' | prepend: site.baseurl }}"><img src="./images/triplea_promo.png" alt="Screenshot of TripleA" width="1059" height="798"></a>
  </div>
+
+<h2> Software Licensing </h2>
+
+<p>
+TripleA is released under the [GNU General Public License](https://www.gnu.org/licenses/old-licenses/gpl-2.0.html) (GPL); either version 2
```
ssoloff: @DanVanAtta Sorry, missed this during the review...  This is a Markdown link, not an HTML link.

ssoloff: #289

PR #289, opened 2018-01-23
Fix broken GPL link

ssoloff:


ssoloff:
Self-merging this since it's an active problem in production.

PR #293, opened 2018-01-31
Google search result enhancement

RoiEXLab:
See https://developers.google.com/search/docs/data-types/logo

I replaced the triplea logo on the website with the 256x256 version found in the assets repository


RoiEXLab:
Just noticing now, that the image is not 100% the same...
Will try to find one that is the same, just with higher resolution

RoiEXLab:
Ok. I ended up simply upscaling the image to 112x112px to match googles requirements.
Should not be noticeable that much.

PR #297, opened 2018-02-11
Combine help and forums header link. 

DanVanAtta:
They both link to the same page, it's a bit odd. Combining helps make it clear they go to the same website, from there it should be straight forward to scan and find the word help

Before:
before

After:
after

### DIFF
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -6,8 +6,7 @@
       <li class="menu-items"><a href="{{ '/download/' | prepend: site.baseurl }}">Download</a></li>
 
       <li class="menu-items"><a href="{{ '/maps/' | prepend: site.baseurl }}">Maps</a></li>
-      <li class="menu-items"><a href="https://forums.triplea-game.org/category/10/help-questions">Help</a></li>
-      <li class="menu-items"><a href="https://forums.triplea-game.org">Forums</a></li>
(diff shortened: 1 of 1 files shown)


ssoloff:
A bit confused here.  Those links look different, with the former showing all threads for a specific topic (help), while the latter goes to the main page.  What am I missing?

DanVanAtta:
First are two links to different pages in forums. Second is one link to forums.

I was clicking through each link recently, it seemed *really* off that we are indexing the forums, 50% of header links were to a different website. The idea is that we can just reduce to one link, and risk the 'help' being a little bit harder to find. I don't think that will be a big impact, meanwhile it reduces header link cont and makes it less 'weird' that we have multiple links to a 'different' website.

DanVanAtta:
This is newly the case too, before we actually had a help page, so help went to an internal page and forums was a link to that location. On the update to have help go to forums, we then introduced the scenario where multiple header links go to forums.

DanVanAtta:
@prastle  ^^ FYI

prastle:
Fine by me
move it around as you wish but yes we need 1 link!

DanVanAtta:
Done! Regrouped the categories a bit, help and questions is number 2. Also put some things as sub-categories to group further.

prastle:
Looks good Off topic but the other three ladders can be closed.  They shut them down from inactivity.  Only the Revised ladder is still running.

PR #298, opened 2018-02-11
Drop static map page, migrate to dynamic maps page

DanVanAtta:
We got pretty far on a project to show a page with dynamic map listing. I rediscovered it recently when clicking the link on the maps page. It seemed really buried, and even though still primitive, it's still far better than the previous map page.

------------------

The update here is to drop our old legacy maps page and link directly to the new maps page. With some small luck we'll see some further iterations to make the maps page even better. As is, having all the different maps front and center is really good to show the breadth and variety, much stronger than a maps page with a few slightly underwhelming screenshots.

----------------

Before clicking maps on header link would result in old maps page with a buried link to the new maps page:
before

After, we drop the old maps page and instead go directly to the link that was on the old maps page, and we get to:
after


ssoloff:
[Travis is failing](https://travis-ci.org/triplea-game/triplea-game.github.io/builds/340025259#L694) with

```
Link 'http://localhost:4000//maps/' returned code 404
```

Presumably due to the removal of _/maps.md_.

PR #299, opened 2018-02-13
Add license

ssoloff:
@ron-murhammer The content of this repo is currently unlicensed.  Not sure if there is anything that would preclude you, as the owner, from applying a license to the existing files.

AGPL-3.0-or-later chosen for this repo because the content is served over a network.


RoiEXLab:
👍

PR #300, opened 2018-03-05
latest is back to 7621

DanVanAtta:


release_notes.md
```diff
@@ -23,6 +19,8 @@ See the [full list here](https://github.com/triplea-game/triplea/pulls?q=merged%
 * Chat message history navigation keys (#2654)
 * Prompt to download Tutorial map (#2650)
 
+See the [full list here](https://github.com/triplea-game/triplea/pulls?q=merged%3A%3E%3D2018-01-10T18%3A12%3A06Z).
+
 See the [full list here](https://github.com/triplea-game/triplea/pulls?q=merged%3A2017-11-25T20%3A11%3A09Z..2018-01-10T18%3A12%3A06Z).
```
ssoloff: FYI, instead of two separate links, you can generate the correct consolidated link using the [generate_release_changes_url](https://github.com/triplea-game/triplea-game.github.io/blob/master/.github/generate_release_changes_url) script.

DanVanAtta: I like the script, but I wonder if this can't be more automated?
I would like release notes to be really light weight, along with the rest of the release process. It's an interesting history of how far we've come.

On the other hand, I still am not sure about sending average users to that list. Seems we are piggy-backing two different needs to a convenient location. Maybe we should keep the latest version links somewhere else?

PR #318, opened 2019-06-10
Move win32 download to a link

DanVanAtta:
There are more than expected windows 32 bit downloads than we would expect. To discourage users from choosing the wrong windows distro, the win32 download button is converted here to a link to "hide it". This will hopefully make it easier to pick the right distribution version, and if someone is not paying a lot of attention and is clicking the first thing that they see that says "windows", then they'll click the 64 bit windows DL now instead of 'accidently' getting the 32 bit.

Screenshot from 2019-06-10 11-13-53

### DIFF
--- a/download.html
+++ b/download.html
@@ -10,10 +10,6 @@ permalink: /download/
   <img src="../images/operating-systems/windows100.png" alt="Windows Icon">
   <p>TripleA&nbsp;for Windows <strong>(64&nbsp;bit)</strong></p>
 </a>
-<a id="win32" href="https://github.com/triplea-game/triplea/releases/latest" target="_blank" class="button dl-links" onclick="hideShowInstallationText(document.getElementById('windows-install'))">
-  <img src="../images/operating-systems/windows100.png" alt="Windows Icon">
-  <p>TripleA&nbsp;for Windows <strong>(32&nbsp;bit)</strong></p>
-</a>
 <a id="macos" href="https://github.com/triplea-game/triplea/releases/latest" target="_blank" class="button dl-links" onclick="hideShowInstallationText(document.getElementById('mac-install'))">
   <img src="../images/operating-systems/mac100.png" alt="Mac Icon">
   <p>TripleA&nbsp;for Mac</p>
@@ -61,6 +57,11 @@ permalink: /download/
 
 <div>
   <h2>Additional Links</h2>
+
+  <p><a id="win32" href="https://github.com/triplea-game/triplea/releases/latest" target="_blank"
+     onclick="hideShowInstallationText(document.getElementById('windows-install'))">
+    Download TripleA&nbsp;for Windows <strong>32&nbsp;bit</strong>
+  </a></p>
   <p><a href="{{ '/release_notes/' | prepend: site.baseurl }}">Release Notes</a></p>
   <p><a href="https://github.com/triplea-game/triplea/wiki/Verifying-Binaries">Verifying Binaries</a></p>
   <p><a href="{{ '/old_downloads/' | prepend: site.baseurl }}">Previous TripleA Releases</a></p>


download.html
```diff
@@ -10,10 +10,6 @@
   <img src="../images/operating-systems/windows100.png" alt="Windows Icon">
   <p>TripleA&nbsp;for Windows <strong>(64&nbsp;bit)</strong></p>
```
RoiEXLab: Actually I'd remove the (64-bit) part, to treat it as "default" version.

However I can see that some users might assume that the right version gets picked for them automatically and wonder why their version is not really working.

download.html
```diff
@@ -61,6 +57,11 @@ <h2>Linux Installation</h2>
...
 <div>
   <h2>Additional Links</h2>
+
+  <p><a id="win32" href="https://github.com/triplea-game/triplea/releases/latest" target="_blank"
+     onclick="hideShowInstallationText(document.getElementById('windows-install'))">
+    Download TripleA&nbsp;for Windows <strong>32&nbsp;bit</strong>
+  </a></p>
   <p><a href="{{ '/release_notes/' | prepend: site.baseurl }}">Release Notes</a></p>
   <p><a href="https://github.com/triplea-game/triplea/wiki/Verifying-Binaries">Verifying Binaries</a></p>
   <p><a href="{{ '/old_downloads/' | prepend: site.baseurl }}">Previous TripleA Releases</a></p>
```
RoiEXLab: A couple of lines below there's some javascript if else block that does show the DL size for the hovered download entry, explicitly stating "win32", that could be removed.

download.html
DanVanAtta: My thoughts on keeping it in were to help notify a 32 bit user that it is not right. Hopefully they have been exposed to enough 64 vs 32 that they are watching for it.

For the most part web users do not read a page and they click on the first thing that looks right and click back if they are wrong. In that case they did not read the 64 bit part to begin with and clicked on the windows button.

download.html
DanVanAtta: I'm reluctant to remove too much just yet; in case we see a sizable number of 32 bit downloads ongoing after this update, then we'll want to restore to the old image. Merge conflicts could come up if we just rely on revert only. I'm personally okay keeping some tech debt here. Let me know if you really think we should do the cleanup right now.

download.html
RoiEXLab: Yeah I assumed you were going for that, but when I learned a thing it's: Never overestimate the users abilities ^^

DanVanAtta:
@RoiEXLab some javascript to bring the right version to the top would be great. We'd want to be careful the UX was good to still find other kinds of downloads and not necessarily hide everything else.

For example, I think adoptopenjdk does this pretty nicely: https://adoptopenjdk.net/

My 2 cents, this is the right move for now. When we have the JS in place, it'll be good to update further with that. We don't have to move to the best solution immediately, or even at all. Looking at the website traffic, it is anemic. There might be higher ROI efforts.

RoiEXLab:
Well I have some spare time right now, it would take me half an hour to change the code so only a single option is shown by default. So if you'd prefer that instead I could do it real quick.
I was thinking about adoptopenjdk as well, but the biggest restriction is that we have to rely on javascript for everything, so we are a little bit more limited on the possibilities.

DanVanAtta:
IMO I'd merge to take the pressure off and then rebase and work from there for a fast iteration. But really up to you @RoiEXLab

RoiEXLab:
#319
Took me longer than expected, I had to fix all the JS mess I created back in the day.
It's stunning how much my code evolved since then ^^

DanVanAtta:
Good stuff @RoiEXLab 
I hear ya on code style changing. I've noticed the same looking at some of my earlier stuff working on TripleA. 
The updated JS is a good bit cleaner FWIW, I follow it much more easily!

PR #319, opened 2019-06-10
Auto-detect optimal download option

RoiEXLab:
Alternative to #318 
Based on https://stackoverflow.com/a/6866569 and https://stackoverflow.com/a/19883965
If that isn't accurate enough we might want to consider using a library like https://github.com/bestiejs/platform.js
If no platform information is available, or the information available does not match anything the same download website is shown which currently exists, with all 4 options.
The same happens when javascript is disabled.

Obviously I haven't been able to test the code on multiple systems, but I'm going to trust the internet on this one 🤷‍♂ 

Preview:
Preview

### DIFF
--- a/download.html
+++ b/download.html
@@ -69,61 +69,74 @@ permalink: /download/
 </div>
 
 <script>
-const additionalElements = document.getElementsByClassName("show-on-install");
+const versionInformation = {
+  win32: {
+    instructionId: 'windows-install'
+  },
+  win64: {
+    instructionId: 'windows-install'
+  },
+  mac: {
+    instructionId: 'mac-install'
+  },
+  linux: {
+    instructionId: 'linux-install'
+  }
+};
+
+const additionalElements = document.getElementsByClassName('show-on-install');
 function hideShowInstallationText(which) {
-  [].forEach.call(document.getElementsByClassName("installation-instructions"), function(current){
-    if(current !== which){
-      current.style.display = "none";
+  Array.from(document.getElementsByClassName('installation-instructions')).forEach(current => {
+    if (current !== which) {
+      current.style.display = 'none';
     }
   });
-  if(which){
-    which.style.display = "block";
-    [].forEach.call(additionalElements, current => current.style.display = "block");
+
+  if (which) {
+    which.style.display = 'block';
+    Array.from(additionalElements).forEach(current => current.style.display = 'block');
   } else {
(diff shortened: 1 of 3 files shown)


download.html
```diff
@@ -23,6 +23,8 @@
   <p>TripleA&nbsp;for Linux</p>
 </a>
 
+<p>Missing a version? Checkout <a href="https://github.com/triplea-game/triplea/releases/latest">all versions here</a>.</p>
```
DanVanAtta: (1) The wording maybe could be different. Not sure if everyone would think of version as say "v1.9" vs "v1.8". I'd re-word this to "Download for other platforms"

(2) Can we still show a download page? Maybe copy/paste the old version of the DL page and call it the 'all-platform' page. It's intended to avoid users from looking at the github release list, it's overwhelming for some.

RoiEXLab: 1. Agreed
2. Would an "un-hide" button be enough? So that all the remaining installers can be shown if requested.

DanVanAtta:
Demo'ed, I like it. I would make it a bit easier to find other platforms.

I'd consider one of two treatments:
(1) update the link text and have the link go to a download page with the 4 buttons we had
(2 leaning towards this personally)  Change the "other versions" text to "download for other platforms" and then below that have the 3 remaining buttons, but smaller.

Something like this:
Screenshot from 2019-06-10 15-15-15

Once nice benefit to that (2) option is a person does not have to read the link text to find the right version in case they want  a different one, and it avoids the need for a second page.

DanVanAtta:
Requesting link text be updated and the other download options not be from github releases.

download.html
DanVanAtta: A "show installers for all platforms" button could work too. That would be similar to the link going to a different page. Offhand I think it's better to show the options off the bat with the right one up top and center. Up to you at this point IMO, main requirement I think is that the other installers not be from github releases.

RoiEXLab:
Ok then, I'll have a look at it tomorrow (currently 1 AM here ^^)

RoiEXLab:
Ok, revision 2:
Screenshot

The current HTML really needs some work, the download buttons size is currently tied to the underlying image size, that's why I used a simple scale transformation to achieve small buttons. This obviously also shrinks down the text size, but otherwise I'd have to redo how those buttons work, a nice follow-up PR.

RoiEXLab:
@DanVanAtta Feel free to restart the build, looks like one of the sites a link points to is down.

PR #322, opened 2019-08-08
Update release notes with notable PRs since November 2018

DanVanAtta:


DanVanAtta:
@RoiEXLab agree, wasn't sure how final it was given the open GL issue. Essentially it is just pretty cutting edge. This update is not the final list, just getting a bit more caught up.

PR #327, opened 2019-11-27
Add /help/ page with simple redirect

DanVanAtta:
Add a simplistic javascript redirect to send requests to /help
to /user-guide. Should fix broken links to /help and address:
https://github.com/triplea-game/triplea-game.github.io/issues/326

### DIFF
--- a/help.html
+++ b/help.html
@@ -1,10 +1,12 @@
 ---
 layout: page
-title: help-redirect
+title: help
 permalink: /help/
 ---
 
 <script>
   window.location.replace("/user-guide");
 </script>
+
+<a href="/user-guide">This page has moved</a>
 
--- a/help.html
+++ b/help.html
@@ -8,5 +8,4 @@ permalink: /help/
   window.location.replace("/user-guide");
 </script>
 
-<a href="/user-guide">This page has moved</a>
-
+<a href="/user-guide/">This page has moved</a>
--- a/help.html
+++ b/help.html
@@ -1,4 +1,5 @@
 ---
+title: help-redirect
 permalink: /help/
 ---
 
--- a/help.html
+++ b/help.html
@@ -1,0 +1,8 @@
+---
+permalink: /help/
+---
+
+<script>
+  window.location.replace("/user-guide");
+</script>
+
--- a/help.html
+++ b/help.html
@@ -1,4 +1,5 @@
 ---
+layout: page
 title: help-redirect
 permalink: /help/
 ---


help.html
```diff
@@ -0,0 +1,10 @@
+---
+layout: page
+title: help-redirect
+permalink: /help/
+---
+
+<script>
+  window.location.replace("/user-guide");
+</script>
```
RoiEXLab: Hmm you might want to consider using a fallback mechanism if javascript is not enabled.
A simple link, or even a meta refresh tag will do the trick already.

DanVanAtta: Good idea, by any chance do you know how to do a meta refresh in markdown?

RoiEXLab: Well you'll have to use normal html, but this'll require you to create a new template or something.
IIRC there was a way to use layouts with parameters, so there should be no need to have a second almost identical template

DanVanAtta: Hmm, at that rate IMO a link and relying on JS is probably sufficient. It's worth us keeping the overhead low. IIRC, virtually all browsers have JS enabled now, we'd be doing the template work just for the bots, and even then, they've the link they can follow.

DanVanAtta: Thoughts? @RoiEXLab

RoiEXLab: I'm not talking about browser support here, I'm mostly concerned about user using addons like noscript etc. that mostly use a default-disabled setting for individual scripts, so you have to opt-in to use them.
So some sort of indicator "please visit page xyz, redirect didn't work for some reason" would be nice so that people don't just have to look at an empty page if anything goes wrong

DanVanAtta: Agreed, a link like "Page moved to 'user-guides'" would be easy enough.
Small delay here, my jekyll broke, had to spend some time fixing it.

DanVanAtta: @RoiEXLab now updated: 535c241

Again, great suggestion/consideration.

With the link, if the redirect does not do its thing, the page will look like this:
Screenshot from 2019-11-27 18-12-26

DanVanAtta:
@RoiEXLab latest build failure could be a problem with the link checker.

I double checked the link and it works from a local jekyll, and of course it must start with `/`

https://travis-ci.org/triplea-game/triplea-game.github.io/builds/617999244?utm_medium=notification&utm_source=email

> Warning: File '/home/travis/build/triplea-game/triplea-game.github.io/_site/help/index.html' had issues: Link 'http://localhost:4000//user-guide' was redirected permanently to 'http://localhost:4000/user-guide/'. Consider updating this link

DanVanAtta:
>  everything too way longer than anticipated

Familiar story :smiley: 
No worries, thank you for the efforts/work @RoiEXLab

RoiEXLab:
Ok, this issue should be fixed in commit https://github.com/RoiEXLab/link-checker/commit/3c6a8c7dea93e8ce7a21c2ced84c827bd68f1988

DanVanAtta:
Thanks @RoiEXLab , I restarted the PR build for this.

RoiEXLab:
@DanVanAtta Now there's the same issue as in #328

DanVanAtta:
This PR neither created nor fixes those build issues, but they should be fixed out of band. I'm not sure I like a process where external dependencies can break unrelated PRs. The linter check is good as a PR check, to avoid the blocking issue it perhaps should be run against master on a cron job.
The as-is configuration is probably fine, though I'm not sure if we can get the problem fixed and might need to remove the old download links if we can't get them fixed.

RoiEXLab:
To be fair, I don't think we'd really notice when a cron job fails.
We probably wouldn't have realized until now if you hadn't created a PR 🤷‍♂

DanVanAtta:
Generally agree @RoiEXLab 

some good news, the broken links are all within release notes, not real harm to converting those links to plain text and/or just having the release tag without artifact.

I thought the links were coming from our previous 'release' download page: https://triplea-game.org/old_downloads/, but they are not. Looking at that page, another problem, it's getting really out of date! We only have one release on it, and it's from several releases back.

RoiEXLab:
Did your tag removal really take all releases with it?
If yes, how did you find the correct commit for the tag so fast?

I don't think losing the artifacts would really be a problem, but it might help to debug a problem in the future if we can get a more narrow range of commits that could have introduced a regression.

DanVanAtta:
> Did your tag removal really take all releases with it?

The corresponding releases, yes. The 1.8.0.9 was kept and any tags well over a  year ago.

> If yes, how did you find the correct commit for the tag so fast?

You have to pro-actively remove tags from your local repo for them to be gone. The list of names was enough to know which tags to push back to origin. Tags are very similar to branches, hence the performance hit, we've effectively 3k+ branches in TripleA due to the tagging.

> I don't think losing the artifacts would really be a problem, but it might help to debug a problem in the future if we can get a more narrow range of commits that could have introduced a regression.

Installing from artifact would only go so far, Mac and non-Mac code *is* different, so are other variants of which JDK, etc...  Having commit range is useful, fundamentally we have the tags around for the last few releases that anyone is playing. The very old tags and having those releases is probably just for nostalgia/historical reasons at this point, I don't know of any real practical purpose behind them at this point.

help.html
```diff
@@ -0,0 +1,12 @@
...
+layout: page
+title: help
+permalink: /help/
+---
+
+<script>
+  window.location.replace("/user-guide");
+</script>
+
+<a href="/user-guide">This page has moved</a>
```
RoiEXLab:
```suggestion
<a href="/user-guide/">This page has moved</a>
```
Oh and btw. this fixes an actual issue caught by the link checker. `/user-guide` redirects to `/user-guide/` which is a completely uneccessary request to be made from a client perspective

DanVanAtta: :+1:

PR #328, opened 2019-11-30
Add codeclimate markdown lint plugin

DanVanAtta:
Set up code climate config file with third party plugin to do
markdown linting.

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,5 @@
+plugins:
+  # https://docs.codeclimate.com/docs/markdownlint
+  markdownlint:
+    enabled: true
+


RoiEXLab:
Interesting, looks like https://github.com/triplea-game/triplea/issues/5608 indeed did have an impact.
I mean good thing it got caught

DanVanAtta:
@RoiEXLab what was the impact? I've missed it so far.

DanVanAtta:
@RoiEXLab ah, spotted it

DanVanAtta:
I pushed the missing tags, with luck should all be okay now.

DanVanAtta:
Actually, we'll need artifacts to be created, I'm not sure how well that is going to go : (

DanVanAtta:
For follow-up, we may want to add eslint as well:

https://docs.codeclimate.com/docs/eslint 
```
 eslint:
    enabled: true
```

RoiEXLab:
I guess this can be merged?

PR #329, opened 2019-12-26
Format linux install command for easier copier paste

DanVanAtta:


download.html
```diff
@@ -52,7 +52,7 @@ <h2>Mac OS X Installation</h2>
 <div id="linux-install" class="installation-instructions">
   <h2>Linux Installation</h2>
   <p>Once the installer finishes downloading, make it executable (chmod +x ./TripleA_*unix.sh)</p>
```
RoiEXLab: What about this line?

download.html
```diff
@@ -52,7 +52,7 @@ <h2>Mac OS X Installation</h2>
...
-  <p>Now run the installer (./TripleA_...unix.sh)</p>
+  <p>Now run the installer: <code>./TripleA_...unix.sh</code></p>
```
RoiEXLab: Hmm could we change this to `./TripleA_*unix.sh`? Would that work?

PR #332, opened 2020-02-08
Rule book html

DanVanAtta:
Takes a pretty solid first stab at porting the rule-book PDF to HTML

To note, the tactical handbook was attached to rule-book and is being split into its own file. The tactical-handbook does not yet have any images or formatting applied to it.


RoiEXLab:
Interesting approach.
I would've tried to use something like https://github.com/mozilla/pdf.js for this, which is the tool that internally converts any PDF to HTML + JS + CSS in order to display it inside your browser.
However there's some JS magic going on there, I wasn't able to cleanly export the generated HTML to just view it inside a browser correctly.

RoiEXLab:
On a second thought I tried the obvious approach using cloudconvert.com with everything embedded in a single HTML file:
[TripleA_RuleBook.html.zip](https://github.com/triplea-game/triplea-game.github.io/files/4174251/TripleA_RuleBook.zip)
Obviously this is a completely different approach to yours, I assume your goal is to make the rule book *easily* maintainable and not just convert it to a text-based format.
However I do think that using `sed` or other regex replacement tools the generated HTML can probably get simplified by a lot so that manual edits become far more feasible (also the images should of course be extracted from the document, just wanted to know if the tool actually procudes a good result).

user-guide/tactical-handbook/tactical-handbook.md
```diff
@@ -0,0 +1,350 @@
...
+Fig. 5. Japan makes a diversionary landing in Alaska to draw American attention away from Germany.
+
+39
+
+
The Objectives of Defense
+There is a symmetry between the objectives of attack and those of defense. A player’s defensive tactics may be influenced by
+any one or combination of the following reasons:
+
+
```
RoiEXLab: Those are some weird characters.
Looks like they are used to display non-standard characters

DanVanAtta: I'm guessing it's the literal bulleted lists.

This iteration still marks the HTML as a WIP, but at least we can iterate on it more. So far this is just a few complete passes:
1. title links (some are actually broken)
2. images for the title sections and they are in place
3. images for the rule book.

We need to still take a couple more passes:
- clean up page numbers
- fix up line breaking
- see about getting some of the images to properly float left/right
- fix up table of content links
- consider updating anything that is out of date, perhaps pruning some items (EG: section on 'undo' buttons).

It seems as well the rule book has a bit of a identity crisis where it's both a how-to manual for how to play the game (mechanically), and it's a rule-book. We may want to consider splitting it up and/or moving the how-to pieces into the tutorial.

DanVanAtta:
Not sure if you saw the forum post that spawned this work as well, for reference @RoiEXLab : https://forums.triplea-game.org/topic/1761/french-translation

RoiEXLab:
I saw the topic but only followed it for the first couple of posts, thanks for the reference.
I already figured it might have something to do with it.

RoiEXLab:
FYI: As you already mentioned some of the links are broken.
The link checker noticed as well, which is the reason the master build is failing

DanVanAtta:
The WIP state is not ideal; we'll get it fixed though.

Didn't mention, I was surprised at how nice a rendering the HTML converter had. Though keeping the text easily maintained is worth quite a bit.

PR #357, opened 2020-07-26
Update lobby-rules.md

prastle:
Updated the active mod list and added the new rules for Private hosts.

### DIFF
--- a/user-guide/lobby-rules.md
+++ b/user-guide/lobby-rules.md
@@ -4,82 +4,37 @@ title: Lobby Rules
 permalink: /user-guide/lobby-rules
 ---
 
+# Lobby Rules
 
-### General Rules
+These general rules apply to the Lobby and any games played in the lobby.
 
-#### Use the golden rule, "Treat others as you would want to be treated.”
-
-We have 5 rules written as follows, all players, including admins, must follow these simple 
-rules when playing on  the lobby, in bots, and when using the forums;
+## Use the golden rule, "Treat others as you would want to be treated."
 
 * Follow the Golden Rule and treat others as you would like to be treated!
 
-* No racism, sexism, and homophobic, or otherwise, vulgar language will be tolerated. 
-The admins do not care if it does not offend you and/or your friends. No bullying and/or 
-discrimination will be tolerated, that includes; Race, Religion, Sex, Age, Cultural Background, 
-Sexual Orientation. Use of foul language, discrimination and derogatory comments are not acceptable, 
-and are subject to admin disciplinary action. It is up to the admin to decide what is or is not 
-acceptable in regards to defamation, discrimination and spamming.
+* No verbal abuse. 
 
-* Due to the recent spat of BLATANT! racism in the lobby the TripleA Council has decided to increase 
-the punishment for these offences to an instant 2 week Ban on first offense!
+* No racist, sexist, homophobic, or otherwise, vulgar language. The moderators
+do not care if it does not offend you and/or your friends. No bullying and/or
+discrimination will be tolerated. It is up to the moderators to decide what is
+or is not acceptable language.
 
-* No verbal abuse. Doing so violates rule #1 above.
+* Listen to the moderators and follow their instructions.
 
-* Listen to the admins and follow their instructions. The admins are here to help the community 
-that you are playing on for free! Do what they say, and if you have a complaint, do it in a civil 
-manner on the forums here. Disobeying an admin can result in you being disciplined by the admins, 
-up to and including being kicked from the game, or temporarily/permanently being banned. We have a 
-thread to lodge complaints, so use it and keep it out of the Lobby.
-
-* Please watch your language - there are underage children playing in the lobby, would you want 
-your younger brother/sister/cousin being exposed to the kind of language you're using? Keep it 
-clean when you get upset, take a breather before expressing your feelings in the Lobby and please 
-refrain from swearing.
-
-Our general policy in regards to breaking the rules is: Warn, Mute, Kick, Day-Ban, Perm Ban. 
-Depending on the seriousness of the issue, you may be insta-day-banned (remember rule #2.) 
-Furthermore, repetitive bad behavior can result in a perm ban, whether its twice or 10 times, 
-it is up to the admins.
+* Please watch your language - Keep it clean when you get upset, take a breather
+before expressing your feelings in the Lobby and please refrain from swearing.
 
 
-### Banning Information
+## Banning Information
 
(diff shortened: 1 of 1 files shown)


user-guide/lobby-rules.md
```diff
@@ -104,3 +104,32 @@ If a user has been found to directly target a person in a server, they will be r
 Remember our Golden Rule is to treat others the same way you would want to be treated. TripleA 
 is one of the greatest Axis and Allies games and community. We want everyone to enjoy their
  time here, so please keep it clean and have fun!
+
```
DanVanAtta, line 107: Why is the list of active moderators useful?
Is this just another thing for us to maintain? When we add moderators, or remove them, do we really want to *have* to update this list?

prastle: Do we leave the Mods and Admins hidden?  I think it should be general knowledge to all users.  Just my two cents.

DanVanAtta: The moderators are marked as such in the chatter list. In that sense it is not hidden. I'm not sure if having the list of moderators be here also solves any issues of a player being confused as to who is a moderator.

prastle:
I replied in our pm. in the past we had "fake" mods hopefully that is easily dealt with now thus u are correct

DanVanAtta:
Prastle, you've raised a concern of 'fake' moderator tags. 
- that should not be that possible. We don't allow spaces.
- the fix for that is not by creating a 'true' list somewhere else. That means players *should* ignore the mod tags and always look to the list. In that case it's argument for removing the mod tag entirely. I think that sounds outlandish, just as requiring players to head out of the application to the user-guide to see who is a mod.

For 'fake' moderators, the fix is with the game engine IMO, not by creating a new list that we have to maintain and hope that players will consult.

user-guide/lobby-rules.md
```diff
@@ -50,11 +50,11 @@ please go to this thread. Banned players: Please keep personal feelings and atti
...
 user, banning admin, or another admin involved, we ask that you do not post unnecessarily.
 
-### Lobby and Bot Rules
+### Lobby, Bot and Private Host Rules
 
-Again the General Rules above apply to the Lobby as well to the bots. The bots are not private 
-rooms and the Admin are also moderating the bots. Admins have the right to take action upon you, 
-if you are found in breach of these rules.
+Again the General Rules above apply to the Lobby as well to the bots and private hosts.
+If you are using the website the Admins have the right to take action upon you.
```
DanVanAtta, line 56: Nitpick, lobby, bot and private hosts are not a website. Perhaps we would be better off to explain that breaches of the rules can result in disciplinary actions including being banned from TripleA.

prastle: Np edit away it was just a general concept.  I do agree that if peeps can fool the lobby display it is an engine problem.  Ill just keep a list of all mods in a new cat on the forum.  Edit as you wish

prastle: @DanVanAtta  Since you asked in our chat I will list a few examples which have been discussed over the years.  I have no problem with it being public since it is common knowledge to most TripleA users.  Cernel would be a good example.  Cernel = CerneI =Cerne! =Cernel1  mainly please note the i is not defined from the l.  Additional points could be made about the mod brackets.  This has happened repeatedly in the past.  Just read the bunker or old git issues.  Thus why I think a list should be maintained.  Just my 2 cents do as you wish. Pras

DanVanAtta: @prastle I'm not aware of either old bunker posts or git issues mentioning the problem of moderator spoofing. Perhaps if you have any mind you could PM bunker links to me?

DanVanAtta:
@prastle I pushed an edit to the rules. A good chunk of the information in the rules is stated twice or over-emphasized. Let me know what you think of the edits (please leave comments on any line numbers where you have comments/questions).

user-guide/lobby-rules.md
prastle: I don't keep track of it.  I do know it was mentioned in the hated link of old during the original 1.9 release but really you would just have to ask Cernel.  he can give you first hand info.  Either way I am done with this since you apparently need proof of any statement.  I already explained exactly what the player did.

prastle:
I only changed 1 line as I recall possible 2 about private hosts and added the list of mods.  Do as you wish.  @DanVanAtta   Apparently you have rewritten a few now. :)

user-guide/lobby-rules.md
DanVanAtta: Just wanting background information @prastle as it might be available.

DanVanAtta:
@prastle please have a look at the updates and let me know if there are any you disagree with or think could be improved. This is  process where we gather concensus before making updates. This way when we see the rules it's not just the dictate of one person and then subject to equally arbitrary change, but is a decision reached by the team (majority decision is effectively anything that any two people agree on).

user-guide/lobby-rules.md
```diff
@@ -4,82 +4,37 @@ title: Lobby Rules
...
-discrimination will be tolerated, that includes; Race, Religion, Sex, Age, Cultural Background, 
-Sexual Orientation. Use of foul language, discrimination and derogatory comments are not acceptable, 
-and are subject to admin disciplinary action. It is up to the admin to decide what is or is not 
-acceptable in regards to defamation, discrimination and spamming.
-
-* Due to the recent spat of BLATANT! racism in the lobby the TripleA Council has decided to increase 
-the punishment for these offences to an instant 2 week Ban on first offense!
-
-* No verbal abuse. Doing so violates rule #1 above.
+* No verbal abuse.
```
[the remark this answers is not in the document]
prastle: cool

user-guide/lobby-rules.md
```diff
@@ -4,82 +4,37 @@ title: Lobby Rules
...
-
-Excessive language, discrimination or openly defacing Lobby admins, Forum mods, and other 
-users can result in being IP banned.
-
-Donations and other favors do not exempt you from following the lobby rules, and does not 
-limit or prevent the mods from taking any actions, including mutes or bans, against you.
-
-### Usernames Rules
+please go to [this thread](https://forums.triplea-game.org/topic/15/been-banned-complaint-or-problem-post-here).
+Banned players: Please keep personal feelings and attitudes out of your
```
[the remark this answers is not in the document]
prastle: nice thinning

DanVanAtta:
Cool, let's keep the conversation going out-of-band from here on accomplishing the goals of having the active mod list (so users can verify the mods they see truly are mods). Thanks for reviewing the additional updates and for getting this ball rolling @prastle  :+1:

PR #365, opened 2020-12-25
Remove-prerelease notice and map version map pages

DanVanAtta:
Remove 'pre-release' notice, the 1.9 version referenced is quite
old now.

The 'version' listed is the notification version and is not very
significant to users. It is also confusing compared to the version of
the map itself. To simplify, it is being removed.


RoiEXLab:
Ah, already forgot about this one.
Back when this feature was new this notice made sense, but now it's good to see it gone

PR #366, opened 2020-12-25
Delete old_downloads page

DanVanAtta:
- The page is getting out of date
- Release notes page has all of the same links and is maintained


Cernelius:
I'd change the "Release Notes" to something like "**All Releases Notes & Available Past Releases**". I actually didn't realize you could get the old releases from there too and that is sometimes very useful (like the recent problem that the export game doesn't work).

DanVanAtta:
I was thinking the same, that the link needed a rename. I went with 'release notes and older downloads'

PR #378, opened 2022-10-10
Added content in sitemap and uploaded the xml version of sitemap

ojasaklechayt:
Please check if these are the needs of the two issues which are to convert sitemap.txt into the sitemap.xml and to add some more content innit.


RoiEXLab:
@ojasaklechayt Thank you for your contribution 👍🏼 

@tvleavitt Can you comment on this?

sitemap.txt
```diff
@@ -1,5 +1,8 @@
-https://triplea-game.github.io/
-https://triplea-game.github.io/download/
-https://triplea-game.github.io/maps-list/maps/
-https://triplea-game.github.io/release_notes/
-https://triplea-game.github.io/user-guide/
+http://triplea-game.github.io/
```
DanVanAtta, line 1: why the change from https to http?

PR #412, opened 2026-04-26
Copilot: Add copilot helper files (instructions + ignore)

DanVanAtta:


.copilotignore
```diff
@@ -0,0 +1,30 @@
+# --- Inherited from .gitignore ---
```
DanVanAtta:
```suggestion
# --- From .gitignore ---
```

PR #413, opened 2026-06-14
Draft release notes for v2026-06

DanVanAtta:


DanVanAtta:
The release notes could use more work.. but something is better than nothing right now

